PR #1, opened 2024-06-06
chore: Initial commit with RAC and UFC DTOs

felipecsl:
## Summary
Extracted DTO classes from https://github.com/Eppo-exp/java-server-sdk and https://github.com/Eppo-exp/android-sdk into this shared library.

Classes are separated into packages `cloud.eppo.rac` and `cloud.eppo.ufc`.

We're using https://jitpack.io/ for deployments for now, until we get Maven Central sorted out. I contacted their support to get access.

## TODOs

- [x] Verify that deserialization works for RAC
- [x] Make sure all tests pass both on jvm and android sdks
- [x] Verify that deserialization works for UFC

## For the future
* Remove duplicate jackson/gson JSON parsing dependencies (nice to have)
* Remode dependency on commons-codec (nice to have)
* Make sure we can publish this to Maven Central and Sonatype snapshots (work in progress)


aarsilv (approved):
Love that we're moving towards shared Java code!

Approving as-is, however I wonder if we shouldn't bother including all the RAC stuff in here. New versions of the Java SDK _should_ be UFC-only. This would keep the code simpler and the size smaller isolating RAC stuff to just the previously released SDKs.

felipecsl:
> Love that we're moving towards shared Java code!
> 
> Approving as-is, however I wonder if we shouldn't bother including all the RAC stuff in here. New versions of the Java SDK _should_ be UFC-only. This would keep the code simpler and the size smaller isolating RAC stuff to just the previously released SDKs.

I think there's no downside of doing so. Even if we don't plan to ever support RAC on Android, it's in general a good idea to have more code sharing, as there are utility functions used by both code paths (eg base64 decoding) which benefit from being colocated in the same repo

felipecsl:
Code is now live at https://s01.oss.sonatype.org/content/repositories/snapshots/cloud/eppo/sdk-common-jvm/ 😎

felipecsl:
I'm gonna merge this and keep iterating on both SDKs separately, thanks for the review!

PR #3, opened 2024-06-10
chore: Upload artifacts from Github CI

felipecsl:
# Description

This PR makes it so every push to `main` will publish a new SNAPSHOT version of the library to Sonatype. This will be the bleeding edge version.

Configured all secrets and verified that the workflow works as intended. Release showed up here https://s01.oss.sonatype.org/content/repositories/snapshots/cloud/eppo/sdk-common-jvm/1.0.0-SNAPSHOT/

### DIFF
--- a/.github/workflows/publish-snapshot.yml
+++ b/.github/workflows/publish-snapshot.yml
@@ -1,4 +1,4 @@
-name: Publish snapshot SDK to the oss.sonatype.org
+name: Publish SDK Snapshot artifact
 
 on:
   pull_request:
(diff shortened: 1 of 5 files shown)


.github/workflows/publish-snapshot.yml
```diff
@@ -0,0 +1,31 @@
+name: Publish SDK Snapshot artifact
+
+on:
+  push:
+    branches: [main]
```
leoromanovsky, line 5: this is great!

leoromanovsky (approved):

PR #4, opened 2024-06-10
chore: Prepare for next release iteration

felipecsl:
Just some housekeeping. Set the version to v1.0.0 so that will be the first non-snapshot version of this artifact. Then we can update both SDKs to use that instead of the snapshot version

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -107,7 +107,7 @@ publishing {
 
 signing {
   sign publishing.publications.mavenJava
-  if (project.hasProperty('CI')) {
+  if (System.env['CI']) {
     useInMemoryPgpKeys(System.env.GPG_PRIVATE_KEY, System.env.GPG_PASSPHRASE)
   }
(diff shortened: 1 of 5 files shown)


felipecsl:
Alright v1.0.0 has been released to maven central, it might take some hours for it to show up

build.gradle
```diff
@@ -6,7 +6,7 @@ plugins {
 }
 
 group = 'cloud.eppo'
-version = '1.0.0-SNAPSHOT'
+version = '2.0.0-SNAPSHOT'
```
felipecsl, line 9: i've already released v1.0.0 so setting this to the potential next version snapshot

felipecsl:
Code is live at https://repo1.maven.org/maven2/cloud/eppo/sdk-common-jvm/1.0.0/

aarsilv (approved):

PR #7, opened 2024-06-11
chore: Remove dependency on Gson

felipecsl:
# Description

We had a duplicate JSON library dependency on both Jackson and Gson.
This change consolidates all json serialization/deserialization into Jackson, removing Gson completely.
Also made some models immutable for good measure. All tests still pass. Moved some tests from the android-sdk codebase into here since they belong here


src/main/java/cloud/eppo/ufc/dto/FlagConfig.java
```diff
@@ -4,63 +4,49 @@
...
-
-  private Map<String, Variation> variations;
-
-  private List<Allocation> allocations;
+  private final String key;
+  private final boolean enabled;
+  private final int totalShards;
+  private final VariationType variationType;
+  private final Map<String, Variation> variations;
+  private final List<Allocation> allocations;
```
aarsilv, line 12: Nice job migrating the DTOs to be immutable 💪

aarsilv (approved):
Nice! Approving this as its certainly one way to go. I do know one customer was interested in keeping dependencies down for the android build, so there was talk of consolidating to what Android uses, `org.json` vs `com.fasterxml.jackson`. Any particular reason for one or the other?

PR #12, opened 2024-06-12
chore: Build and publish artifacts with Java 8

felipecsl:
This makes sure that this library remains java 8 compatible, otherwise it can't be use from https://github.com/Eppo-exp/java-server-sdk

### DIFF
--- a/.github/workflows/lint-test-sdk.yml
+++ b/.github/workflows/lint-test-sdk.yml
@@ -4,6 +4,8 @@ on:
   pull_request:
     paths:
       - '**/*'
+  push:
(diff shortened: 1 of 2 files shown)


.github/workflows/lint-test-sdk.yml
```diff
@@ -4,6 +4,8 @@ on:
   pull_request:
     paths:
       - '**/*'
+  push:
+    branches: [main]
```
felipecsl, line 8: run lint and test on every push to main

leoromanovsky (approved):

PR #19, opened 2024-07-02
fix(ci): Add missing checkout fetch depth for spotless, version bump

felipecsl:
Should fix https://github.com/Eppo-exp/sdk-common-jdk/actions/runs/9765392722/job/26955946224

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -6,7 +6,7 @@ plugins {
 }
 
 group = 'cloud.eppo'
-version = '2.0.0'
+version = '2.1.0-SNAPSHOT'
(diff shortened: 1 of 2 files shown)


build.gradle
```diff
@@ -6,7 +6,7 @@ plugins {
 }
 
 group = 'cloud.eppo'
-version = '2.0.0'
+version = '2.1.0-SNAPSHOT'
```
felipecsl, line 9: I've already published v2 locally, so bumping the version here already

leoromanovsky (approved):

PR #25, opened 2024-08-06
Add bandit support

aarsilv:
_Eppo Internal:_
🎟️ **Ticket:** [FF-2743 - Common Shared Bandits Tests Passing](https://linear.app/eppo/issue/FF-2743/common-shared-bandits-tests-passing)

This pull request adds support for contextual multi-armed bandits to the common SDK for future use by upstream Universal Flag Configuration (UFC)-powered Java and--perhaps in the future--Android SDKs.

Much of the classes, tests, and logic have been ported from the existing Java bandits implementation, updated to account for new decisions and edge-case handling as bandit development has taken place.

Bandit functionality is verified against our [shared SDK test data](https://github.com/Eppo-exp/sdk-test-data/tree/main) bandit tests. Additional tests ensure bandits log and handle errors as expected.

image


src/main/java/cloud/eppo/ConfigurationStore.java
```diff
@@ -1,53 +1,106 @@
...
       log.warn("Flags missing in configuration response");
       flags = new ConcurrentHashMap<>();
+      banditReferences = new ConcurrentHashMap<>();
     } else {
       // TODO: atomic flags to prevent clobbering like android does
       // Record that flags were set from a response so we don't later clobber them with a
       // slow cache read
       flags = new ConcurrentHashMap<>(config.getFlags());
-      log.debug("Loaded " + flags.size() + " flags from configuration response");
+      banditReferences = new ConcurrentHashMap<>(config.getBanditReferences());
```
aarsilv, line 47: `banditReferences` is a top-level UFC property containing some information on any bandits at play (but not their model data, which is provided separately)

src/main/java/cloud/eppo/ConfigurationStore.java
```diff
@@ -1,53 +1,106 @@
...
   }
+
+  public String banditKeyForVariation(String flagKey, String variationValue) {
+    // Note: In practice this double loop should be quite quick as the number of bandits and bandit
+    // variations will be small
+    //       Should this ever change, we can optimize things.
+    for (Map.Entry<String, BanditReference> banditEntry : banditReferences.entrySet()) {
+      BanditReference banditReference = banditEntry.getValue();
+      for (BanditFlagVariation banditFlagVariation : banditReference.getFlagVariations()) {
+        if (banditFlagVariation.getFlagKey().equals(flagKey)
```
aarsilv, line 68: While we _could_ use the shortcut of assuming the variation key and value and bandit key are all the same, since the UFC explicitly provides the information to be more specific, I'll take advantage of it. We can have other SDKs do this once needed.

src/main/java/cloud/eppo/EppoClient.java
```diff
@@ -26,12 +25,14 @@ public class EppoClient {
...
 
+  private final ConfigurationStore configurationStore;
   private final ConfigurationRequestor requestor;
   private final AssignmentLogger assignmentLogger;
-  private final boolean isGracefulMode;
+  private final BanditLogger banditLogger;
   private final String sdkName;
   private final String sdkVersion;
   private final boolean isConfigObfuscated;
+  private boolean isGracefulMode;
```
aarsilv, line 35: made non-final so that graceful mode can be toggled at runtime. Useful for tests, and possibly for developers too.

src/main/java/cloud/eppo/EppoClient.java
```diff
@@ -40,22 +41,21 @@ public class EppoClient {
...
 
-  /** @noinspection FieldMayBeFinal */
-  private static ConfigurationStore configurationStoreOverride = null;
-
   private EppoClient(
       String apiKey,
       String sdkName,
       String sdkVersion,
       String host,
+      ConfigurationStore configurationStore,
```
aarsilv, line 49: Configuration store now passed in (this is similar to other SDKs and hopefully paves the way for easier caching in Android and whatnot)

src/main/java/cloud/eppo/BanditEvaluationResult.java
```diff
@@ -0,0 +1,73 @@
+package cloud.eppo;
+
+import cloud.eppo.ufc.dto.DiscriminableAttributes;
+
+public class BanditEvaluationResult {
+
+  private final String flagKey;
+  private final String subjectKey;
+  private final DiscriminableAttributes subjectAttributes;
```
aarsilv, line 9: `DiscriminableAttributes` is an interface I introduced for attributes that can broken up (discriminated) by numeric and contextual. You'll see it more in action later, but basically, it helps abstract converting broken-out attributes to combined and vice versa.

src/main/java/cloud/eppo/EppoClient.java
```diff
@@ -121,7 +131,7 @@ public void refreshConfiguration() {
   protected EppoValue getTypedAssignment(
       String flagKey,
       String subjectKey,
-      SubjectAttributes subjectAttributes,
+      Attributes attributes,
```
aarsilv: Note the `SubjectAttributes` class has been renamed to `Attributes` as they are used with actions too now

src/main/java/cloud/eppo/EppoClient.java
```diff
@@ -121,7 +131,7 @@ public void refreshConfiguration() {
   protected EppoValue getTypedAssignment(
       String flagKey,
       String subjectKey,
-      SubjectAttributes subjectAttributes,
+      Attributes subjectAttributes,
```
aarsilv, line 134: Note the `SubjectAttributes` class has been renamed to `Attributes` as they are used with actions too now

src/main/java/cloud/eppo/FlagEvaluationResult.java
```diff
@@ -1,72 +1,61 @@
...
 public class FlagEvaluationResult {
 
-  private String flagKey;
-  private String subjectKey;
-  private SubjectAttributes subjectAttributes;
-  private String allocationKey;
-  private Variation variation;
-  private Map<String, String> extraLogging;
-  private boolean doLog;
+  private final String flagKey;
```
aarsilv, line 9: Switched to the immutable result pattern used elsewhere

src/main/java/cloud/eppo/Utils.java
```diff
@@ -40,7 +40,7 @@ public static String getMD5Hex(String input) {
     return hashText.toString();
   }
 
-  public static Date parseUtcISODateElement(JsonNode isoDateStringElement) {
+  public static Date parseUtcISODateNode(JsonNode isoDateStringElement) {
```
aarsilv, line 43: method name didn't change in our switch from org.json to Jackson, so updating to keep things parallel

src/main/java/cloud/eppo/logging/Assignment.java
```diff
@@ -1,64 +1,40 @@
...
     this.allocation = allocation;
     this.variation = variation;
     this.subject = subject;
-    this.timestamp = timestamp;
     this.subjectAttributes = subjectAttributes;
     this.extraLogging = extraLogging;
     this.metaData = metaData;
   }
 
-  public static Assignment createWithCurrentDate(
```
aarsilv, line 41: seemed superfluous 🪓

src/main/java/cloud/eppo/logging/Assignment.java
```diff
@@ -79,11 +55,11 @@ public String getSubject() {
     return subject;
   }
 
-  public String getTimestamp() {
+  public Date getTimestamp() {
```
aarsilv, line 58: Moving toward the pattern of providing native Date/Time objects for the loggers, and then they can log to their warehouse however they want (vs. us being opinionated by providing an ISO string)

src/main/java/cloud/eppo/ufc/dto/Attributes.java
```diff
@@ -0,0 +1,64 @@
+package cloud.eppo.ufc.dto;
+
+import java.util.HashMap;
+import java.util.Map;
+import java.util.stream.Collectors;
+
+public class Attributes extends HashMap<String, EppoValue> implements DiscriminableAttributes {
```
aarsilv, line 7: This is a rename from `SubjectAttributes`, but we've added enough new stuff that Git is confused.

This and `ContextAttributes` implement `DiscriminableAttributes`, the difference is the former has the numeric vs. categorical type explicitly specified.

src/main/java/cloud/eppo/ufc/dto/BanditReference.java
```diff
@@ -0,0 +1,21 @@
+package cloud.eppo.ufc.dto;
+
+import java.util.List;
+
+public class BanditReference {
```
aarsilv, line 5: New UFC feature 📈

src/main/java/cloud/eppo/ufc/dto/DiscriminableAttributes.java
```diff
@@ -0,0 +1,10 @@
+package cloud.eppo.ufc.dto;
+
+public interface DiscriminableAttributes {
```
aarsilv, line 3: Created this interface as some parts of bandit evaluation don't care about numeric vs. categorical (i.e., _variation_ assignment) while others do (i.e., _action_ assignment). This interface:
1. Allows upstream developers to easily pass in explicitly separated attributes or not as they please.
2. Simplifies the bandit code that interacts with the attributes

src/main/java/cloud/eppo/ufc/dto/EppoValue.java
```diff
@@ -107,4 +108,25 @@ public String toString() {
             "Cannot stringify Eppo Value type " + this.type.name());
     }
   }
+
+  @Override
+  public boolean equals(Object otherObject) {
```
aarsilv, line 113: Overriding `equals()` and `hashCode()` allows for the comparison of different instances of Eppo values based on the value they contain. This is particularly useful for `assertEquals()` in automated tests.

src/main/java/cloud/eppo/ufc/dto/adapters/BanditParametersResponseDeserializer.java
```diff
@@ -11,21 +11,38 @@
...
   }
 
-  protected BanditsDeserializer(Class<?> vc) {
+  protected BanditParametersResponseDeserializer(Class<?> vc) {
     super(vc);
   }
 
   @Override
-  public Map<String, BanditParameters> deserialize(
+  public BanditParametersResponse deserialize(
```
aarsilv, line 32: We now return the whole response instead of just handling the `bandits` property of it to make this file more analogous to `FlagConfigResponseDeserializer`

src/main/java/cloud/eppo/ufc/dto/adapters/FlagConfigResponseDeserializer.java
```diff
@@ -36,41 +35,58 @@ public FlagConfigResponseDeserializer() {
...
+                banditReferences.put(field.getKey(), banditReference);
+              });
     }
 
-    return new FlagConfigResponse(flags);
+    return new FlagConfigResponse(flags, banditReferences);
   }
 
-  private FlagConfig deserializeFlag(JsonNode jsonNode, DeserializationContext context) {
+  private FlagConfig deserializeFlag(JsonNode jsonNode) {
```
aarsilv, line 79: `DeserializationContext` is not needed in these sub functions

src/test/java/cloud/eppo/EppoClientBanditTest.java
```diff
@@ -0,0 +1,336 @@
...
+      "dummy-bandits-api-key"; // Will load bandit-flags-v1
+  private static final String TEST_HOST =
+      "https://us-central1-eppo-qa.cloudfunctions.net/serveGitHubRacTestFile";
+  private static final ObjectMapper mapper = new ObjectMapper().registerModule(module());
+
+  private static final AssignmentLogger mockAssignmentLogger = mock(AssignmentLogger.class);
+  private static final BanditLogger mockBanditLogger = mock(BanditLogger.class);
+  private static final Date testStart = new Date();
+
+  // TODO: possibly consolidate code between this and the non-bandit test
```
aarsilv, line 42: Thinking of leaving this as a TODO for now until everything is said and done. I don't want to prematurely combine parts of tests only to have things diverge once non-bandit-only (for now) things like caching is layered on.

src/test/java/cloud/eppo/EppoClientBanditTest.java
```diff
@@ -0,0 +1,336 @@
...
+  }
+
+  @BeforeEach
+  public void reset() {
+    clearInvocations(mockAssignmentLogger);
+    clearInvocations(mockBanditLogger);
+    EppoClient.getInstance().setIsGracefulFailureMode(false);
+  }
+
+  @ParameterizedTest
```
aarsilv, line 67: Shared UFC Bandit Tests! 💪

src/test/java/cloud/eppo/EppoClientTest.java
```diff
@@ -1,28 +1,23 @@
 package cloud.eppo;
 
-import static org.junit.Assert.assertEquals;
-import static org.junit.Assert.assertFalse;
-import static org.junit.Assert.assertNotNull;
-import static org.junit.Assert.assertThrows;
-import static org.junit.Assert.assertTrue;
-import static org.junit.Assert.fail;
+import static org.junit.jupiter.api.Assertions.assertEquals;
```
aarsilv, line 3: Consolidating on the jupiter flavors to match the imported `@Test`

src/test/java/cloud/eppo/EppoClientTest.java
```diff
@@ -201,26 +192,26 @@ private <T> void assertAssignment(
 
     if (assignment instanceof Boolean) {
       assertEquals(
-          failureMessage, expectedSubjectAssignment.getAssignment().booleanValue(), assignment);
+          expectedSubjectAssignment.getAssignment().booleanValue(), assignment, failureMessage);
```
aarsilv, line 195: jupiter flavored asserts have the error message at the end (which I like more anyways 🤷‍♂️ )

src/test/java/cloud/eppo/helpers/BanditTestCase.java
```diff
@@ -0,0 +1,36 @@
+package cloud.eppo.helpers;
+
+import java.util.List;
+
+public class BanditTestCase {
+  private final String flag;
+  private final String defaultValue;
+  private final List<SubjectBanditAssignment> subjects;
+  private String fileName;
```
aarsilv, line 9: Set later (after deserialization) to make helpful test case "names" that are displayed in IDE/terminal

src/test/resources/bandits-parameters-1.json
```diff
@@ -0,0 +1,101 @@
+{
```
aarsilv, line 1: For now, our deserializer tests use their own files (as opposed to the shared UFC test file). I think this is ok, as it makes the deserializer tests less fragile. The shared test data _should_ ensure that the SDK correctly handles anything new thrown at it. Nevertheless, we could consider having the deserializer tests also use shared test data.

src/test/java/cloud/eppo/EppoClientTest.java
```diff
@@ -390,18 +362,9 @@ public void testAssignmentEventCorrectlyCreated() {
...
-      dateFormat.setTimeZone(java.util.TimeZone.getTimeZone("UTC"));
-      Date parsedTimestamp = dateFormat.parse(capturedAssignment.getTimestamp());
-      assertNotNull(parsedTimestamp);
-      assertTrue(parsedTimestamp.after(testStart));
-      assertTrue(parsedTimestamp.before(assertionDate));
-    } catch (ParseException ex) {
-      throw new RuntimeException(ex);
-    }
+    assertTrue(capturedAssignment.getTimestamp().after(testStart));
+    Date inTheNearFuture = new Date(System.currentTimeMillis() + 1);
```
aarsilv, line 366: Deflake the test for when things are particularly speedy

src/main/java/cloud/eppo/ConfigurationStore.java
```diff
@@ -1,53 +1,105 @@
...
     } else if (flags.isEmpty()) {
-      log.warn("Request for flag " + flagKey + " with empty flags");
+      log.warn("Request for flag {} with empty flags", flagKey);
     }
     return flags.get(flagKey);
   }
+
+  public String banditKeyForVariation(String flagKey, String variationValue) {
+    // Note: In practice this double loop should be quite quick as the number of bandits and bandit
+    // variations will be small. Should this ever change, we can optimize things.
```
giorgiomartini0, line 64: Agreed - no need to prematurely optimize this.

giorgiomartini0 (approved):
Impeccable! Big THANK YOU for the detailed comments explaining what changed and why 🙏

src/test/java/cloud/eppo/helpers/AssignmentTestCaseDeserializer.java
```diff
@@ -60,7 +60,7 @@ private List<SubjectAssignment> deserializeSubjectAssignments(JsonNode jsonNode)
   }
 
   private TestCaseValue deserializeTestCaseValue(JsonNode jsonNode) {
-    if (jsonNode != null && jsonNode.isObject()) {
+    if (jsonNode != null && (jsonNode.isObject() || jsonNode.isArray())) {
```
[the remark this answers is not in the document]
aarsilv: We have a test case with JSON array now! https://github.com/Eppo-exp/sdk-test-data/blob/main/ufc/tests/test-json-array-config-flag.json#L4

Added four days ago ([PR](https://github.com/Eppo-exp/sdk-test-data/pull/50)) when a customer caught it wasn't working for them 😞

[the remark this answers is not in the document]
typotter: I actually [rolled back that test case](https://github.com/Eppo-exp/sdk-test-data/pull/51) earlier after talking with Leo. Some of the SDKs would need a fair amount of work to support the JSON Array of Objects as a variation value. More work, I think, than is agreeable while it's questionable whether support for this is desired.

src/main/java/cloud/eppo/BanditEvaluator.java
```diff
@@ -0,0 +1,167 @@
...
+    for (Map.Entry<String, Double> actionScore : actionScores.entrySet()) {
+      if (actionScore.getKey().equals(highestScoredAction)) {
+        // The highest scored action is weighed at the end
+        continue;
+      }
+
+      // Compute weight (probability)
+      double unboundedProbability =
+          1 / (actionScores.size() + (gamma * (highestScore - actionScore.getValue())));
+      double boundedProbability = Math.max(unboundedProbability, actionProbabilityFloor);
```
typotter, line 124: In the other SDKs, we divide [actionProbabilityFloor by the number of actions to weigh;](https://github.com/Eppo-exp/python-sdk/blob/70595d1c00851f383095c777de6480bbb882b480/eppo_client/bandit.py#L167)

[Comment 🧵 ](https://github.com/Eppo-exp/python-sdk/pull/38/files#r1619789521)

src/main/java/cloud/eppo/BanditEvaluator.java
```diff
@@ -0,0 +1,167 @@
...
+    // all be shifted to the same new action at the same time.
+    List<String> shuffledActionKeys =
+        actionWeights.keySet().stream()
+            .sorted(
+                Comparator.comparingInt(
+                        (String actionKey) ->
+                            ShardUtils.getShard(
+                                flagKey + "-" + subjectKey + "-" + actionKey,
+                                BANDIT_ASSIGNMENT_SHARDS))
+                    .thenComparing(actionKey -> actionKey))
```
typotter, line 149: 👍

src/main/java/cloud/eppo/ConfigurationRequestor.java
```diff
@@ -12,28 +14,40 @@ public class ConfigurationRequestor {
...
 
+  // TODO: async loading for android
   public void load() {
     log.debug("Fetching configuration");
-    Response response = client.get("/api/flag-config/v1/config");
+    String flagConfigurationJsonString = requestBody("/api/flag-config/v1/config");
+    configurationStore.setFlagsFromJsonString(flagConfigurationJsonString);
+
+    Set<String> neededModelVersions = configurationStore.banditModelVersions();
+    boolean needBanditParameters = !loadedBanditModelVersions.containsAll(neededModelVersions);
```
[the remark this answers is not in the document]
typotter: Shooting for bleeding edge all at once eh? I like your style 🤠

src/main/java/cloud/eppo/ConfigurationRequestor.java
```diff
@@ -12,28 +14,44 @@ public class ConfigurationRequestor {
...
     this.client = client;
+    this.loadedBanditModelVersions = new HashSet<>();
   }
 
+  // TODO: async loading for android
   public void load() {
     log.debug("Fetching configuration");
-    Response response = client.get("/api/flag-config/v1/config");
+    String flagConfigurationJsonString = requestBody("/api/flag-config/v1/config");
+    configurationStore.setFlagsFromJsonString(flagConfigurationJsonString);
```
typotter, line 29: Do you need to worry about threading here and lock reads until the bandit models are set below?

src/main/java/cloud/eppo/ConfigurationRequestor.java
```diff
@@ -12,28 +14,44 @@ public class ConfigurationRequestor {
...
+
+    Set<String> neededModelVersions = configurationStore.banditModelVersions();
```
typotter, line 31: Is it possible for more than one bandit to have the same model version string?

src/main/java/cloud/eppo/ConfigurationStore.java
typotter: [😅 ](https://github.com/Eppo-exp/php-sdk/pull/35/files#diff-8b2fba0ae27745d3578b6cd1a5f13a6edef0672eeaca2404b159d93b4b7cc82dR52)

src/main/java/cloud/eppo/EppoClient.java
```diff
@@ -412,9 +409,67 @@ private JsonNode parseJsonString(String jsonString) {
...
+                  subjectAttributes.getCategoricalAttributes(),
+                  banditResult.getActionAttributes().getNumericAttributes(),
+                  banditResult.getActionAttributes().getCategoricalAttributes(),
+                  buildLogMetaData());
+
+          banditLogger.logBanditAssignment(banditAssignment);
+        }
+
+        // Update result to reflect that we've been assigned an action
+        result = new BanditResult(assignedVariation, banditResult.getActionKey());
```
typotter, line 454: result should get set above the call to `banditLogger.logBanditAssignment`. This ensures that if the user-provided bandit logger throws an exception, the `throwIfNotGraceful` call below has an accurate result to return.

src/main/java/cloud/eppo/logging/BanditAssignment.java
```diff
@@ -0,0 +1,101 @@
+package cloud.eppo.logging;
+
+import cloud.eppo.ufc.dto.Attributes;
+import java.util.Date;
+import java.util.Map;
+
+public class BanditAssignment {
```
typotter, line 7: Would you consider _public final_ fields instead of _getters_? I'm not sure what the java wisdom would say here.

src/main/java/cloud/eppo/ufc/dto/BanditActions.java
```diff
@@ -0,0 +1,20 @@
+package cloud.eppo.ufc.dto;
+
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Set;
+import java.util.stream.Collectors;
+
+public class BanditActions extends HashMap<String, DiscriminableAttributes> implements Actions {
```
typotter, line 8: 🧑‍🍳

typotter (changes requested):
what a monster of a pull 💪 
Requesting changes but only because of the [math mismatch here](https://github.com/Eppo-exp/sdk-common-jdk/pull/25/files#diff-4a070310188bb5ae2af93410324cfd699da47ddc18ef9f92c18ba7f3fdc8581fR124)

src/test/java/cloud/eppo/helpers/AssignmentTestCaseDeserializer.java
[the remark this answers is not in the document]
aarsilv: Rolling back the test case sounds good. Any harm in leaving support in SDKs that can handle it (this one, node, etc.?)

src/main/java/cloud/eppo/BanditEvaluator.java
aarsilv: Yikes! GREAT catch 🎉  Dividing by actions is what we should do. 

This logic was part of what was copied over from the original Java implementation, before we changed up our strategy, and I failed to update it.

What this does mean is that we don't have a test enforcing that logic. It may be hard to put one together, but I'll see if I can. 🤔

src/test/java/cloud/eppo/helpers/AssignmentTestCaseDeserializer.java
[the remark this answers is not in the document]
typotter: no hard that I can think of - especially on the test parsing side.

src/main/java/cloud/eppo/BanditEvaluator.java
typotter: The reason why this doesn't get teased out in a test is because we're using small min probability floors and typically a gamma of 1 with only a couple of actions. With some tweaking of scoring parameters, should be able to create a test case that pushes it over the edge

src/main/java/cloud/eppo/ConfigurationRequestor.java
aarsilv: Good thinking! The underlying implementation is a thread-safe ConcurrentHashMap, which is what we are using in the Android SDK without (known) issues

src/main/java/cloud/eppo/ConfigurationRequestor.java
aarsilv: No, model versions are unique to each bandit.

Under the hood, this version ID--for now--is the auto incrementing primary key in a model_versions table that has versions for _all_ bandits (as it includes the bandit id)

src/main/java/cloud/eppo/EppoClient.java
aarsilv: Great point. That should be done _and_ all logger calls should be wrapped in try-catches so they are non fatal

src/main/java/cloud/eppo/logging/BanditAssignment.java
aarsilv: My impression is that Java has a weird obsession with getters 😅  The way you access fields for `Record`s seems to confirm this.

Avoiding directly referencing fields will also make interacting with this class consistent with all the others, which I like.

src/main/java/cloud/eppo/BanditEvaluator.java
aarsilv: PR for test cases that cover this (and more!) here: https://github.com/Eppo-exp/sdk-test-data/pull/52

build.gradle
```diff
@@ -140,14 +141,15 @@ tasks.withType(PublishToMavenRepository) {
   }
 }
 
-signing {
-  sign publishing.publications.mavenJava
-  if (System.env['CI']) {
-    useInMemoryPgpKeys(System.env.GPG_PRIVATE_KEY, System.env.GPG_PASSPHRASE)
+if (!project.gradle.startParameter.taskNames.contains('publishToMavenLocal')) {
```
aarsilv, line 144: Don't sign the maven package if publishing to local maven repository (for multi-repository development)

src/main/java/cloud/eppo/EppoClient.java
```diff
@@ -412,9 +409,71 @@ private JsonNode parseJsonString(String jsonString) {
...
+
+      String banditKey = configurationStore.banditKeyForVariation(flagKey, assignedVariation);
+      if (banditKey != null && !actions.isEmpty()) {
+        BanditParameters banditParameters = configurationStore.getBanditParameters(banditKey);
+        BanditEvaluationResult banditResult =
+            BanditEvaluator.evaluateBandit(
+                flagKey, subjectKey, subjectAttributes, actions, banditParameters.getModelData());
+
+        // Update result to reflect that we've been assigned an action
+        result = new BanditResult(assignedVariation, banditResult.getActionKey());
```
aarsilv, line 435: @typotter we now set the action part of the result as soon as we get it, and surround bandit logging with a `try`-`catch` like we do assignment logging.

Non-fatal logging error tests were added as well.

src/test/java/cloud/eppo/EppoClientTest.java
```diff
@@ -411,6 +374,50 @@ public void testAssignmentEventCorrectlyCreated() {
     assertEquals(expectedMeta, capturedAssignment.getMetaData());
   }
 
+  @Test
+  public void testAssignmentLogErrorNonFatal() {
```
aarsilv, line 378: Adding a test to ensure errors with logging assignments are non-fatal

src/test/java/cloud/eppo/EppoClientBanditTest.java
```diff
@@ -0,0 +1,358 @@
...
+          EppoClient.getInstance()
+              .getBanditAction(
+                  "banner_bandit_flag", "subject", new Attributes(), actions, "default");
+      assertEquals("banner_bandit", banditResult.getVariation());
+      assertNull(banditResult.getAction());
+    }
+  }
+
+  @Test
+  public void testBanditLogErrorNonFatal() {
```
aarsilv, line 334: @typotter test case to ensure any issues with the user-provided bandit logging method do not interfere with returning the bandit result.

src/main/java/cloud/eppo/BanditEvaluator.java
```diff
@@ -0,0 +1,168 @@
...
+    for (Map.Entry<String, Double> actionScore : actionScores.entrySet()) {
+      if (actionScore.getKey().equals(highestScoredAction)) {
+        // The highest scored action is weighed at the end
+        continue;
+      }
+
+      // Compute weight (probability)
+      double unboundedProbability =
+          1 / (actionScores.size() + (gamma * (highestScore - actionScore.getValue())));
+      double minimumProbability = actionProbabilityFloor / actionScores.size();
```
aarsilv, line 124: @typotter bandit probability floor now normalized

src/test/resources/bandits-parameters-1.json
typotter: just my 0.02 (or 0.0144USD after adjusting for ForEx)
For deserialization, I think using these files is the best way to go. Imo, it doesn't make a lot of sense to pull in dynamic data, parse it, and then assert on hard-coded values. Regardless of whether the underlying data is stable, the mismatch of dynamic vs coded feels awkward and hangs a lampshade on the fragility.

typotter (approved):
⛵

PR #30, opened 2024-08-15
Optimize assignment getting

aarsilv:
_Eppo Internal_:
🎟️ **Ticket:** [FF-3037 - Optimize Java SDK getMD5Hex() to not use BigInteger](https://linear.app/eppo/issue/FF-3037/optimize-java-sdk-getmd5hex-to-not-use-biginteger)
🗨️ **Slack:** ["...they're specifically calling out the java sdk using biginteger calculations which is taking up most of the overhead..."](https://eppo-group.slack.com/archives/C04KR1GGE3C/p1723562511849529?thread_ts=1723561977.939239&cid=C04KR1GGE3C)

Profiling our Java SDK, there were a few inefficiencies in the way we computed the MD5 hash as a String, and the way we used hashing to generate an integer within a shard space:
* Hasing used `BigInteger`, which was inefficient
* Shard space generating used the string hash to then compute a number, which is an unnecessary middle step

This PR refactors `getMD5hash()` to not use `BigInteger`, and `getShard()` to not use `getMD5Hash()`.

It puts in place a test that generates 10,000 assignments and asserts that it computes quickly. Using that test, it seems these optimizations generate about a 6x speed up.


build.gradle
```diff
@@ -17,6 +17,7 @@ dependencies {
   // For UFC DTOs
   implementation 'commons-codec:commons-codec:1.17.0'
   implementation 'org.slf4j:slf4j-api:2.0.13'
+  testImplementation 'org.slf4j:slf4j-simple:2.0.16'
```
aarsilv, line 20: Gives our unit test environment a logger so log statements actually show up 😅

src/main/java/cloud/eppo/ShardUtils.java
```diff
@@ -1,18 +0,0 @@
...
-
-import cloud.eppo.model.ShardRange;
-
-public final class ShardUtils {
-  /** Returns whether the given shard is within the range. */
-  public static boolean isShardInRange(int shard, ShardRange range) {
-    return shard >= range.getStart() && shard < range.getEnd();
-  }
-
-  public static int getShard(String input, int maxShardValue) {
```
aarsilv, line 11: THis method was both rewritten and moved to `Utils` to cluster it with the similar `getMD5Hex()`

src/main/java/cloud/eppo/ShardUtils.java
```diff
@@ -1,18 +0,0 @@
...
-    StringBuilder hashText = new StringBuilder(Utils.getMD5Hex(input));
```
aarsilv, line 12: 🪓 The unnecessary middle operation we're removing

src/main/java/cloud/eppo/Utils.java
```diff
@@ -16,28 +15,60 @@
...
   private static final Logger log = LoggerFactory.getLogger(Utils.class);
+  private static final MessageDigest md = buildMd5MessageDigest();
+
+  private static MessageDigest buildMd5MessageDigest() {
+    try {
+      return MessageDigest.getInstance("MD5");
+    } catch (NoSuchAlgorithmException e) {
+      throw new RuntimeException("Error computing md5 hash", e);
+    }
+  }
```
aarsilv, line 26: Instantiate the `MessageDigest` once instead of every call to `getMD5Hex()` or `getShard()`

We use `reset()` to ensure calls don't interfere with each other.

src/test/java/cloud/eppo/ProfileBaseEppoClientTest.java
```diff
@@ -0,0 +1,94 @@
...
+    Map<String, AtomicInteger> variationCounts = new HashMap<>();
+    Attributes subjectAttributes = new Attributes();
+    subjectAttributes.put("country", "FR");
+
+    ThreadMXBean threadBean = ManagementFactory.getThreadMXBean();
+    long startTime = threadBean.getCurrentThreadCpuTime();
+
+    int numIterations = 10000;
+
+    for (int i = 0; i < numIterations; i++) {
```
aarsilv, line 64: These 10,000 iterations take ~70 ms to run on my machine, so this is still a fast (enough) test.

felipecsl (approved):
nice tests

src/test/java/cloud/eppo/UtilsTest.java
```diff
@@ -11,6 +11,34 @@
 import org.junit.jupiter.api.Test;
 
 public class UtilsTest {
+  @Test
+  public void testGetMd5Hash() {
```
aarsilv, line 15: Tests save lives!

src/test/java/cloud/eppo/ProfileBaseEppoClientTest.java
```diff
@@ -0,0 +1,98 @@
...
+    // Expect ~30% control (50% of 60%)
+    assertEquals(0.30, variationCounts.get("control").doubleValue() / numIterations, 0.02);
+    // Expect ~18% red (30% of 60%)
+    assertEquals(0.18, variationCounts.get("red").doubleValue() / numIterations, 0.02);
+    // Expect ~12% yellow (20% of 60%)
+    assertEquals(0.12, variationCounts.get("yellow").doubleValue() / numIterations, 0.02);
+
+    // Seeing ~48,000,000 - ~54,000,000 for 10k iterations on a M2 Macbook Pro; let's fail if more
+    // than 150,000,000; giving a generous allowance for slower systems (like GitHub) but will still
+    // catch if things slow down considerably
```
chasdevs, line 92: slick

typotter: ➕ 1️⃣

PR #32, opened 2024-08-16
Improvements for use by eventual application

aarsilv:
_Eppo Internal:_
🎟️ **Ticket:** [FF-3066 - Improve common SDK for use by customer apps](https://linear.app/eppo/issue/FF-3066/improve-common-sdk-for-use-by-customer-apps)
👯‍♀️ **Related PR:** [`java-bandit-proxy #7`](https://github.com/Eppo-exp/java-bandit-proxy/pull/7)

To make it so the eventual application using the Java or Android SDK can use interfaces defined in here, their packaging will expose the API of this package. To help make it clear which classes are actually used via exposure, we've created an `api` package and put the relevant interfaces and classes in there.

We also create a new method on `Attributes` that serialize the instance to a JSON string. This is very useful when logging, as people will want to save these as JSON to their data warehouses. (Note: we are ressurecting a flavor of this from the old Java SDK).

Also, to ensure compatibility with Java 8 we set the compile target.

### DIFF
--- a/src/main/java/cloud/eppo/api/Attributes.java
+++ b/src/main/java/cloud/eppo/api/Attributes.java
@@ -1,5 +1,8 @@
-package cloud.eppo.ufc.dto;
+package cloud.eppo.api;
 
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.databind.node.ObjectNode;
 import java.util.HashMap;
 import java.util.Map;
 import java.util.stream.Collectors;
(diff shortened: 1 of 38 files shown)


src/main/java/cloud/eppo/api/Attributes.java
```diff
@@ -61,4 +64,43 @@ public Attributes getCategoricalAttributes() {
...
+          continue;
+        }
+        if (attributeValue.isBoolean()) {
+          result.put(attributeName, attributeValue.booleanValue());
+          continue;
+        }
+        // fall back put treating any other eppo values as a string
+        result.put(attributeName, attributeValue.toString());
+      }
+    }
```
felipecsl, line 98: I don't think you need any of this. Jackson supports serializing maps to json, just pass that directly to `writeValueAsString`

aarsilv: I'm not sure this would work, the value of the map is an `EppoValue`. Would Jackson just call `toString()` to get the property value? Does it then check to see whether it needs quotes or not? (e.g., if the EppoValue's `toString()` evaluated to `1` would that be a string or number to serialize? It seems to me it may be better to explicitly control this.

felipecsl: im pretty sure it will just call `toString` for you

felipecsl: it won't handle numbers vs. strings, though. everything will be treated as a string

felipecsl: if you need that you'd have to write a custom deserializer

aarsilv: Yeah and they'd have to then provide that deserializer do their call I think, so thinking we just leave the convenience method? It's also possible they don't even have Jackson at play in their codebase (only our package)

felipecsl: we can have the deserializer as part of the library that we just add to the ObjectMapper on line 75, but this is simple enough that I'm ok keeping as is. if it ends up growing in complexity then I think having a serializer makes sense, but LGTM as is

felipecsl (approved):

src/main/java/cloud/eppo/api/Attributes.java
felipecsl: i dont know why I keep saying "deserializer" when I mean "serializer" 🤦🏽

PR #38, opened 2024-09-17
feat: Allow initialization using a JSON string.

typotter:
Eppo internal:
🎟️ towards FF-3245

## Motivation and Context

In order to speed initialization and ultimately to support a _no-network_ mode, the Android SDK, which has not migrated to the latest common package, requires the ability to initialize the `EppoClient` from a JSON string containing the FlagResponse.

## Overview of Changes
- Move all  config data into `Configuration`; an immutable object with a builder to facilitate building a complete configuration (flags and bandits) while conditionally loading bandits.
- The `Configuration` now handles all access to the flag and bandit config, including checking for obfuscated config and hashing the flag key as necessary.
- Add `initialConfiguration` to `init` signature
- Allow `ConfigurationStore` to be constructed with initial configuration values
- Tests

## Testing
- Unit tests


src/main/java/cloud/eppo/ConfigurationStore.java
```diff
@@ -20,9 +20,20 @@ public class ConfigurationStore {
...
+    this(null, null);
+  }
+
+  public ConfigurationStore(String initialFlagConfiguration, String initialBanditParameters) {
     flags = new ConcurrentHashMap<>();
     banditReferences = new ConcurrentHashMap<>();
     banditParameters = new ConcurrentHashMap<>();
+
+    if (initialFlagConfiguration != null) {
+      setFlagsFromJsonString(initialFlagConfiguration);
```
rasendubi, line 32: minor(unrelated to this PR): `ConcurrentHashMap` is used incorrectly in `ConfigurationStore` as they are never modified — only read. It's better to use a regular hashmap here.

major(but again unrelated to the PR): `flags`/`banditReferences` fields are overwritten without any synchronization, so users may see inconsistent values (e.g., bandit references out-of-sync with flags, etc) or with inconsistent ordering.

src/main/java/cloud/eppo/ConfigurationStore.java
```diff
@@ -20,9 +20,20 @@ public class ConfigurationStore {
   private Map<String, BanditParameters> banditParameters;
 
   public ConfigurationStore() {
+    this(null, null);
+  }
+
+  public ConfigurationStore(String initialFlagConfiguration, String initialBanditParameters) {
```
rasendubi, line 26: It probably needs to store obfuscation flag as well. Otherwise, it's likely susceptible to a bug: when it is initialized with non-obfuscated config and fetches an obfuscated config later on, for a short duration it might either interpret old configuration as obfuscated, or interpret the new configuration as non-obfuscated, leading to evaluation errors

build.gradle
```diff
@@ -6,7 +6,7 @@ plugins {
 }
 
 group = 'cloud.eppo'
-version = '3.0.3-SNAPSHOT'
+version = '3.0.4-SNAPSHOT'
```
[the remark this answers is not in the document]
typotter: I wasn't sure about whether to bump minor now or when cutting a release. I had planned to wrap up some core configuration refactoring into this package before releasing it.
I went ahead and bumped the minor version now anyway

src/main/java/cloud/eppo/ConfigurationStore.java
typotter: I've moved to a `ConfigurationBuffer` object to encapsulate the both flag and bandit data and the decoding of the config. This moves setting the configuration to a single operation. I've also switched to immutable maps since we're not modifying them.

src/main/java/cloud/eppo/ConfigurationStore.java
typotter: Good catch.
With the move to `ConfigurationBuffer`, I've added a field for `isConfigObfuscated`.
`EppoClient` can be initialized with an obfuscated configuration (or not), independent of whether it should expect obfuscated config from the API server.

typotter:
Thanks @rasendubi . Made some larger changes, most of which are forthcoming in a larger refactor to support Andoid

src/main/java/cloud/eppo/ConfigurationRequester.java
```diff
@@ -51,8 +54,4 @@ private String requestBody(String route) {
       throw new RuntimeException(e);
     }
   }
-
-  public FlagConfig getConfiguration(String flagKey) {
```
typotter, line 55: move all config access to the `ConfigurationStore`

src/main/java/cloud/eppo/configuration/ConfigurationBuffer.java
```diff
@@ -0,0 +1,116 @@
...
+import java.util.Set;
+import java.util.stream.Collectors;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * Wrapper object passed between the various Configuration handling classes encapsulating flag and
+ * bandit configuration as well as whether the config is obfuscated.
+ */
+public class ConfigurationBuffer {
```
rasendubi, line 18: nit: In other SDKs, this is called `Configuration`.

I think that's a simpler name (i.e., there's no buffering in `ConfigurationBuffer`). It's also a part of user-exposed API, so it's good to be consistent (unless there's a reason to differ)

src/main/java/cloud/eppo/configuration/ConfigurationBuffer.java
```diff
@@ -0,0 +1,116 @@
...
+
+  private static final Logger log = LoggerFactory.getLogger(ConfigurationBuffer.class);
+
+  private final Map<String, BanditReference> banditReferences;
+  private final Map<String, FlagConfig> flags;
+  private Map<String, BanditParameters> bandits;
```
rasendubi, line 24: minor: is it hard to make `bandits` final as well? I think it's best to make this object completely immutable, so there's less chance for inconsistencies and synchronization issues

src/main/java/cloud/eppo/BaseEppoClient.java
```diff
@@ -134,7 +151,7 @@ protected EppoValue getTypedAssignment(
 
     FlagEvaluationResult evaluationResult =
         FlagEvaluator.evaluateFlag(
-            flag, flagKey, subjectKey, subjectAttributes, isConfigObfuscated);
+            flag, flagKey, subjectKey, subjectAttributes, configurationStore.isConfigObfuscated());
```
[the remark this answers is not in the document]
typotter: Way better to just pack up all the config into one immutable object, I agree.
Thanks for calling this out - there's a number of places where access to the config was disjointed but now `ConfigurationStore` only deals with getting and setting the whole immutable config object at once.

src/main/java/cloud/eppo/BaseEppoClient.java
```diff
@@ -433,7 +450,7 @@ public BanditResult getBanditAction(
 
   private Map<String, String> buildLogMetaData() {
     HashMap<String, String> metaData = new HashMap<>();
-    metaData.put("obfuscated", Boolean.valueOf(isConfigObfuscated).toString());
+    metaData.put("obfuscated", Boolean.valueOf(configurationStore.isConfigObfuscated()).toString());
```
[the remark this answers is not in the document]
typotter: 🛠️ fixed

src/main/java/cloud/eppo/configuration/ConfigurationBuffer.java
typotter: Agreed.  
Now that I've reoriented around the Configuration as the main source of truth and there's a builder, it's no longer buffering the config values from User => configStore or ConfigRequester => configStore

src/main/java/cloud/eppo/configuration/ConfigurationBuffer.java
typotter: it was.. and also wasn't. Moving to a Builder adds a few lines but makes building immutable Configuration objects pretty clean.

Since we want to conditionally load bandits, we need to parse the flag config and compare against some set of bandit models. Moving that workload into the `Configuration.Builder` now eliminates another race (previously, the flag config was set into the config store, which parsed and provided the bandit models needed to be loaded, thus potentially having an invalid config until the bandit response was set).

src/main/java/cloud/eppo/Configuration.java
```diff
@@ -0,0 +1,203 @@
...
+      Map<String, BanditReference> banditReferences,
+      Map<String, BanditParameters> bandits,
+      boolean isConfigObfuscated) {
+    this.flags = flags;
+    this.banditReferences = banditReferences;
+    this.bandits = bandits;
+    this.isConfigObfuscated = isConfigObfuscated;
+  }
+
+  public FlagConfig getFlag(String flagKey) {
```
typotter, line 66: Most of the methods here for parsing and retrieving are cut&paste from `ConfigurationStore`

src/main/java/cloud/eppo/ConfigurationRequester.java
```diff
@@ -0,0 +1,50 @@
+package cloud.eppo;
```
typotter, line 1: Not sure why git is no longer pick up on the filename change

build.gradle
```diff
@@ -6,7 +6,7 @@ plugins {
 }
 
 group = 'cloud.eppo'
-version = '3.0.3-SNAPSHOT'
+version = '3.1.0-SNAPSHOT'
```
aarsilv, line 9: 👍

src/main/java/cloud/eppo/BaseEppoClient.java
```diff
@@ -49,6 +48,28 @@ protected BaseEppoClient(
...
+  protected BaseEppoClient(
+      String apiKey,
+      String sdkName,
+      String sdkVersion,
+      String host,
+      AssignmentLogger assignmentLogger,
+      BanditLogger banditLogger,
+      boolean isGracefulMode,
+      boolean expectObfuscatedConfig,
+      Configuration initialConfiguration) {
```
aarsilv, line 72: Wondering if we should call this class `FlagConfiguration` and parameter `initialFlagConfiguration`, just to disambiguate from other types of configuration (e.g., the config of the SDK with API key, name, version)

Update: ok I see now have we `ConfigurationStore` and whatnot and configuration applies to flags so I think we're good after all 👌

src/main/java/cloud/eppo/BaseEppoClient.java
```diff
@@ -62,16 +83,16 @@ protected BaseEppoClient(
     }
 
     EppoHttpClient httpClient = buildHttpClient(host, apiKey, sdkName, sdkVersion);
-    this.configurationStore = new ConfigurationStore();
-    requestor = new ConfigurationRequestor(configurationStore, httpClient);
+    this.configurationStore = new ConfigurationStore(initialConfiguration);
```
aarsilv, line 86: 🔥

src/main/java/cloud/eppo/BaseEppoClient.java
```diff
@@ -62,16 +83,16 @@ protected BaseEppoClient(
...
+    requester = new ConfigurationRequester(configurationStore, httpClient);
```
aarsilv, line 87: So my IDE suggests this spelling & rename as well, and I'm fine with changing, but the reason I hadn't was to keep this in parallel with our other SDKs which all also use requestor 🤷‍♂️

src/main/java/cloud/eppo/BaseEppoClient.java
```diff
@@ -62,16 +83,16 @@ protected BaseEppoClient(
...
     this.assignmentLogger = assignmentLogger;
     this.banditLogger = banditLogger;
     this.isGracefulMode = isGracefulMode;
     // Save SDK name and version to include in logger metadata
     this.sdkName = sdkName;
     this.sdkVersion = sdkVersion;
     // For now, the configuration is only obfuscated for Android clients
-    this.isConfigObfuscated = expectObfuscatedConfig;
+    this.expectObfuscatedConfig = expectObfuscatedConfig;
```
aarsilv, line 95: good rename

src/main/java/cloud/eppo/BaseEppoClient.java
```diff
@@ -431,7 +450,7 @@ public BanditResult getBanditAction(
     }
   }
 
-  private Map<String, String> buildLogMetaData() {
+  private Map<String, String> buildLogMetaData(boolean isConfigObfuscated) {
```
aarsilv, line 453: Are we passing this through because it's possible the initial configuration may be passed differently than whether it expects obfuscated configurations? I'm ok with this but having trouble thinking when the two would be out of sync and wondering if this is adding complexity vs just using what is passed into the constructor.

src/main/java/cloud/eppo/Configuration.java
```diff
@@ -0,0 +1,203 @@
...
+      // Build the flags config from the json string.
+      FlagConfigResponse config;
+      try {
+        config = mapper.readValue(flagJson, FlagConfigResponse.class);
+      } catch (JsonProcessingException e) {
+        throw new RuntimeException(e);
+      }
+
+      if (config == null || config.getFlags() == null) {
+        log.warn("Flags missing in ufc response");
```
aarsilv, line 144: Thinking we adjust this error message since this could also be called with provided JSON

src/main/java/cloud/eppo/Configuration.java
```diff
@@ -0,0 +1,203 @@
...
+        flags = Collections.emptyMap();
+      } else {
+        flags = Collections.unmodifiableMap(config.getFlags());
```
aarsilv, line 147: nice

src/main/java/cloud/eppo/Configuration.java
```diff
@@ -0,0 +1,203 @@
...
+          .collect(Collectors.toSet());
+    }
+
+    public Set<String> referencedBanditModelVersion() {
+      return banditReferences.values().stream()
+          .map(BanditReference::getModelVersion)
+          .collect(Collectors.toSet());
+    }
+
+    public Builder banditParametersFrom(Configuration currentConfig) {
```
aarsilv, line 170: `banditParametersFromConfig`?

src/main/java/cloud/eppo/Configuration.java
```diff
@@ -0,0 +1,203 @@
...
+      }
+      return this;
+    }
+
+    public Builder banditParameters(String banditParameterJson) {
+      BanditParametersResponse config;
+      try {
+        config = mapper.readValue(banditParameterJson, BanditParametersResponse.class);
+      } catch (JsonProcessingException e) {
+        log.error("Unable to parse bandit parameters response");
```
aarsilv, line 184: here too we may want to weed out the word "response"

src/main/java/cloud/eppo/ConfigurationRequester.java
aarsilv: weird!

src/test/java/cloud/eppo/BaseEppoClientBanditTest.java
```diff
@@ -287,4 +307,100 @@ public void testBanditLogErrorNonFatal() {
         ArgumentCaptor.forClass(BanditAssignment.class);
     verify(mockBanditLogger, times(1)).logBanditAssignment(banditLogCaptor.capture());
   }
+
+  @Test
+  public void testWithInitialConfiguration() {
+    String flagConfig =
+        "{\"flags\": {\n"
```
aarsilv, line 314: rather than define this string like this, consider putting this in the `src/test/resources` directory as a `.json` file and read it in for the tests. This would make working with this much easier (at the downside of it being located in a different file). Or you could even just re-use the existing files in that directory if they work well enough.

src/test/java/cloud/eppo/BaseEppoClientBanditTest.java
```diff
@@ -287,4 +307,100 @@ public void testBanditLogErrorNonFatal() {
...
+    actions.put("adidas", new Attributes());
+
+    BanditResult result =
+        eppoClient.getBanditAction(
+            "banner_bandit_flag", "subject", new Attributes(), actions, "default");
+
+    assertEquals("banner_bandit", result.getVariation());
+    assertEquals("adidas", result.getAction());
+
+    //     Demonstrate that loaded configuration is different from the initial string passed above.
```
aarsilv, line 398: fancy!

aarsilv (approved):
Awesome stuff! Love the way this is coming together and the helpful cleanup and reactors along the way. Approving with a slew of comments you are welcome to take or leave as you see fit!

src/main/java/cloud/eppo/BaseEppoClient.java
rasendubi: yeah, `Configuration` also includes bandit models and maybe more in the future

src/main/java/cloud/eppo/BaseEppoClient.java
rasendubi: If initial configuration is not obfuscated, there's going to be a change of obfuscation'ness on Android after the configuration is changed from the server. Because this may happen concurrently mid-evaluation, evaluation must hold and use a stable `Configuration` reference that doesn't change under it

src/main/java/cloud/eppo/ConfigurationRequester.java
rasendubi: Git data model doesn't have a notion of renames (in git commit, it's one file deleted and another one added). So whenever a file is deleted, git applies a heuristic to find the file if it was renamed (and maybe slightly modified). With more changes added, it looks like we're past git's modification threshold for it to be considered a rename

src/main/java/cloud/eppo/BaseEppoClient.java
```diff
@@ -91,7 +112,7 @@ private EppoHttpClient buildHttpClient(
   }
 
   protected void loadConfiguration() {
-    requestor.load();
+    requester.load(expectObfuscatedConfig);
```
rasendubi, line 115: nit: it seems like we could pass `expectObfuscatedConfig` to requester constructor. The expectation never changes in runtime and it's not used anywhere else beside the requester.

src/main/java/cloud/eppo/BaseEppoClient.java
```diff
@@ -384,6 +402,7 @@ public BanditResult getBanditAction(
       Actions actions,
       String defaultValue) {
     BanditResult result = new BanditResult(defaultValue, null);
+    final Configuration config = configurationStore.getConfiguration();
     try {
       String assignedVariation =
           getStringAssignment(
```
rasendubi, line 408: 🐛 (but not introduced in this PR): `getStringAssignment` might get different `Configuration`, so `banditKeyForVariation` might fail unexpectedly. "obfuscated" in metadata might also be incorrect for this reason. This is an existing/old bug

src/main/java/cloud/eppo/Configuration.java
```diff
@@ -0,0 +1,203 @@
...
+    private Map<String, BanditParameters> bandits = Collections.emptyMap();
+
+    public static Configuration emptyConfig() {
+      return new Configuration(
+          Collections.emptyMap(), Collections.emptyMap(), Collections.emptyMap(), false);
+    }
+
+    public Builder(String flagJson) {
+      this(flagJson, false);
+    }
```
rasendubi, line 125: minor: I would say there's no good default value for `isConfigObfuscated`, so I would rather remove this constructor.
```suggestion
```

src/main/java/cloud/eppo/Configuration.java
rasendubi, line 147: minor: I suggest that we store the whole config/bandit responses "as is". There's a couple of other fields that we'll need in the future. e.g., evaluation reasons need environment name and config creation timestamp

In the future, this will also make it easier serializing `Configuration` to a JSON if we want to pass it on to another client (e.g., frontend initialized from backend response)

src/main/java/cloud/eppo/ConfigurationRequester.java
```diff
@@ -0,0 +1,50 @@
...
+    configurationStore.setConfiguration(configBuilder.build());
+  }
+
+  private String requestBody(String route) {
+    Response response = client.get(route);
+    if (!response.isSuccessful() || response.body() == null) {
+      throw new RuntimeException("Failed to fetch from " + route);
+    }
+    try {
+      return response.body().string();
```
rasendubi, line 45: minor(perf, unrelated to this PR): we could avoid converting bytes to string as JSON parser accepts bytes as input

src/main/java/cloud/eppo/ConfigurationStore.java
```diff
@@ -1,105 +1,24 @@
...
-    // variations will be small. Should this ever change, we can optimize things.
-    for (Map.Entry<String, BanditReference> banditEntry : banditReferences.entrySet()) {
-      BanditReference banditReference = banditEntry.getValue();
-      for (BanditFlagVariation banditFlagVariation : banditReference.getFlagVariations()) {
-        if (banditFlagVariation.getFlagKey().equals(flagKey)
-            && banditFlagVariation.getVariationValue().equals(variationValue)) {
-          return banditEntry.getKey();
-        }
-      }
+      configuration = Configuration.Builder.emptyConfig();
```
rasendubi, line 13: super nit: having this method on `Configuration` could be cleaner:
```suggestion
      configuration = Configuration.empty();
```

rasendubi (approved):

src/main/java/cloud/eppo/Configuration.java
typotter: ✔️

src/main/java/cloud/eppo/Configuration.java
typotter: This is a really good nit :). "response" is like 98% correct here because we're parsing a string into a `BanditParametersResponse` object, but the string may not be a "response" per se and we want to convey a local context to the dev/logs and in that local context it's a JSON string. 🧑‍🍳

src/test/java/cloud/eppo/BaseEppoClientBanditTest.java
typotter: ✔️

src/main/java/cloud/eppo/BaseEppoClient.java
typotter: I had thought to do this but missed it in the last pass. thanks!

src/main/java/cloud/eppo/BaseEppoClient.java
typotter: worth addressing. I filed an issue in linear

src/main/java/cloud/eppo/Configuration.java
typotter: agreed

src/main/java/cloud/eppo/Configuration.java
typotter: ✔️ capturing the JSON string where parsing is successful

src/main/java/cloud/eppo/ConfigurationRequester.java
typotter: done. Storing the raw configs as byte[] and converting from string where needed.

src/main/java/cloud/eppo/ConfigurationStore.java
typotter: agreed.

src/main/java/cloud/eppo/ConfigurationStore.java
```diff
@@ -1,68 +1,52 @@
...
-    }
+  private ConfigurationBuffer configurationBuffer;
 
-    if (config == null || config.getFlags() == null) {
-      log.warn("Flags missing in configuration response");
-      flags = new ConcurrentHashMap<>();
-      banditReferences = new ConcurrentHashMap<>();
+  public ConfigurationStore(final ConfigurationBuffer initialConfiguration) {
+    if (initialConfiguration != null) {
+      this.configurationBuffer = initialConfiguration;
```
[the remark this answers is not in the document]
typotter: 🛠️

PR #39, opened 2024-09-20
feat: Async configuration loading with Cached config fallbacks

typotter:
## Description of changes
Generally, the changes can be followed through the data handling classes; `EppoHttpClient` -> `ConfigurationRequestor` -> `ConfigurationWranger` -> `BaseEppoClient` and `IConfigurationStore` -> `ConfigurationStore` -> `ConfigurationWrangler`.

- The callback interface for `EppoHttpClient` is moved to that class as an inner declaration (this pattern is followed at each layer).
- `EppoHttpClient` returns bytes (directly via `load()` and async via `load(callback)`) instead of `String` or `Response`. `byte[]` is sufficient for the JSON parser so converting to `String` is an extra step.
- `Configuration` is moved to the `api` package as it is exposed to callers of the `BaseEppoClient` via the `initialConfiguration` parameter.
- `ConfigurationRequestor` only requests and returns `Configuration` through sync (pre-existing) and a new `loadAsync` method. `ConfigurationRequestor` sheds the reference to `ConfigurationStore` and passes the responsibility to `ConfigurationWrangler`
- `ConfigurationWrangler` happily yanks code from [The android SDK](https://github.com/Eppo-exp/android-sdk/blob/d33df2918ad7b6b906f637abb67cc81054b89465/eppo/src/main/java/cloud/eppo/android/ConfigurationRequestor.java#L24) to handle the race condition between two async providers of Configuration data, and wrangler the desired values by way of a precedence strategy (currently first-to-return when `loadAsync` is called)
- `ConfigurationWrangler.load` waits for the API response to return and if it fails, falls back to the `ConfigurationStore`
-`loadConfigurationAsync` method added to `BaseEppoClient`


src/main/java/cloud/eppo/BaseEppoClient.java
```diff
@@ -111,10 +118,31 @@ private EppoHttpClient buildHttpClient(
...
+   * Loads the configuration asynchronously, off-thread
+   *
+   * @param skipCache This allows the Configuration Wrangler to skip trying to load from the cache.
+   *     This is useful when we want only the freshest configuration.
+   * @param callback Called when load is complete.
+   */
+  protected void loadConfigurationAsync(boolean skipCache, InitializationCallback callback) {
+    wrangler.loadAsync(
+        skipCache,
+        new ConfigurationWrangler.LoadCallback() {
```
typotter, line 134: these two callbacks actually have the same shape.
`InitializationCallback`, however, is part of this package's protected API via this method, so it inclined me to separate the two.

src/main/java/cloud/eppo/BaseEppoClient.java
```diff
@@ -111,10 +126,36 @@ private EppoHttpClient buildHttpClient(
...
+   *
+   * @param skipCache This allows the Configuration Wrangler to skip trying to load from the cache.
+   *     This is useful when we want only the freshest configuration.
+   * @param callback Called when load is complete.
+   */
+  protected void loadConfigurationAsync(
+      boolean skipCache, @Nullable InitializationCallback callback) {
+    configurationWrangler.loadAsync(
+        skipCache,
+        new ConfigurationWrangler.LoadCallback() {
```
typotter, line 143: these two callbacks actually have the same shape.
`InitializationCallback`, however, is part of this package's protected API via this method, so it inclined me to separate the two.

src/main/java/cloud/eppo/BaseEppoClient.java
```diff
@@ -111,10 +129,36 @@ private EppoHttpClient buildHttpClient(
...
```
typotter, line 146: these two callbacks actually have the same shape.
`InitializationCallback`, however, is part of this package's protected API via this method, so it inclined me to separate the two.

src/main/java/cloud/eppo/EppoHttpClient.java
```diff
@@ -41,34 +41,22 @@ private static OkHttpClient buildOkHttpClient() {
...
-            .addQueryParameter("sdkName", sdkName)
-            .addQueryParameter("sdkVersion", sdkVersion)
-            .build();
+  public byte[] get(String path) {
+    Request request = buildRequest(path);
 
-    Request request = new Request.Builder().url(httpUrl).build();
-    try {
-      return client.newCall(request).execute();
+    try (Response response = client.newCall(request).execute()) {
```
typotter, line 47: _try-with-resources_ block. Modern java is neat.

src/main/java/cloud/eppo/EppoHttpClient.java
```diff
@@ -41,34 +41,22 @@ private static OkHttpClient buildOkHttpClient() {
...
-  // TODO: use this for Java, callback for Android; clean as needed
-  public Response get(String path) {
-    HttpUrl httpUrl =
-        HttpUrl.parse(baseUrl + path)
-            .newBuilder()
-            .addQueryParameter("apiKey", apiKey)
-            .addQueryParameter("sdkName", sdkName)
-            .addQueryParameter("sdkVersion", sdkVersion)
-            .build();
+  public byte[] get(String path) {
```
typotter, line 44: pass around bytes unless a string is absolutely needed. The JSON parser takes bytes, so we save an unneeded conversion step

src/main/java/cloud/eppo/ConfigurationRequestor.java
```diff
@@ -1,55 +1,102 @@
...
 
   public ConfigurationRequestor(
-      ConfigurationStore configurationStore,
-      EppoHttpClient client,
-      boolean expectObfuscatedConfig) {
-    this.configurationStore = configurationStore;
+      EppoHttpClient client, boolean expectObfuscatedConfig, boolean supportBandits) {
     this.client = client;
     this.expectObfuscatedConfig = expectObfuscatedConfig;
+    this.supportBandits = supportBandits;
```
typotter, line 25: This seemed a half-decent way to avoid loading bandit data to the sub-SDKs (android) which don't support bandits at this time. 
If this seems ok, I'll plumb it through the `BaseEppoClient`.

build.gradle
```diff
@@ -6,7 +6,7 @@ plugins {
 }
 
 group = 'cloud.eppo'
-version = '3.1.0-SNAPSHOT'
+version = '3.2.0-SNAPSHOT'
```
typotter, line 9: API modifications:
- `loadConfigurationAsync` method on `BaseEppoClient`
- `Configuration` moved to `api` package.

src/main/java/cloud/eppo/ConfigurationStore.java
```diff
@@ -1,24 +1,23 @@
...
 import org.jetbrains.annotations.NotNull;
 
-public class ConfigurationStore {
+/** In-memory store of a configuration snapshot. */
+public class ConfigurationStore implements IConfigurationStore {
 
-  private volatile Configuration configuration;
+  private volatile Configuration configuration = null;
 
-  public ConfigurationStore(final Configuration initialConfiguration) {
```
typotter, line 9: initial configuration is moved to the wrangler

src/main/java/cloud/eppo/ConfigurationWrangler.java
```diff
@@ -0,0 +1,241 @@
...
+import org.slf4j.LoggerFactory;
+
+public class ConfigurationWrangler {
+  private static final Logger log = LoggerFactory.getLogger(ConfigurationWrangler.class);
+
+  private final ConfigurationRequestor requestor;
+  private final IConfigurationStore configurationStore;
+
+  // This reference is **the** reference to the current configuration.
+  private volatile Configuration config;
```
aarsilv, line 19: Oh man I forgot about the `volatile` keyword!

src/main/java/cloud/eppo/ConfigurationWrangler.java
```diff
@@ -0,0 +1,241 @@
...
+
+  ConfigurationWrangler(
+      ConfigurationRequestor requestor,
+      IConfigurationStore persistentSource,
+      Configuration initialConfiguration) {
+    this.requestor = requestor;
+    this.configurationStore = persistentSource;
+
+    setConfig(initialConfiguration != null ? initialConfiguration : Configuration.emptyConfig());
```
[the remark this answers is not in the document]
aarsilv: In the real world we've seen the CDN beat the device filesystem sometimes 🤯  so we need to handle either one winning properly.

The way the current Android SDK functions is reading and writing to the cache are protected by `synchronized` blocks, and the API writes the configuration after a fetch. For an "offline" initialization--but one perhaps still provided from an external source such as inlined in a server response somewhere--I think its the same situation and we'd want to save it until we have something newer.

src/main/java/cloud/eppo/ConfigurationWrangler.java
```diff
@@ -0,0 +1,241 @@
...
+  }
+
+  private final ReentrantLock lock = new ReentrantLock();
+
+  private void setConfig(Configuration configuration) {
+    lock.lock();
```
aarsilv, line 34: do we need the robustness of a `ReentrantLock` or would `synchronized` do?

src/main/java/cloud/eppo/ConfigurationWrangler.java
```diff
@@ -0,0 +1,241 @@
...
+    try {
+      config = configuration;
+
+      // Fire-and-forget save call to the config store. It may or may not block (non-blocking is
```
aarsilv, line 38: seems worth it to still catch errors and log them

src/main/java/cloud/eppo/ConfigurationWrangler.java
```diff
@@ -0,0 +1,241 @@
...
+    } finally {
+      lock.unlock();
+    }
+  }
+
+  /**
+   * Loads the configuration from the `ConfigurationRequestor` and saves to "persistent" storage
+   * using `IConfigurationStore`
+   *
+   * <p>This method attempts to load from the `IConfigurationStore` if it was unable to load from
```
aarsilv, line 50: ~Our old implementation had the two race, and I think that may still be the approach we want to take~ Ah I see, this is a fallback.

I think we want to take a page from the JavaScript client SDK here (see [it's docs](https://docs.geteppo.com/sdks/client-sdks/javascript/#configuration-cache))

Basically, configuration stors have an `isExpired()` method. If we have a non-expired store, we just use that (no fetch necessary). If it is expired, we fetch. 

A few configuration options dictate behavior. If `useExpired` is true, then it will initialize right away. It will still fetch, and update based on `updateOnFetch` value.

This is all pretty complicated, so if it's too much to mirror the JavaScript client we don't have to. Also the iOS SDK doesn't either. However, consistency there would be good so while messing around this could be the right time to do it.

src/main/java/cloud/eppo/ConfigurationWrangler.java
```diff
@@ -0,0 +1,241 @@
...
+
+            @Override
+            public void onFailure(String errorMessage) {
+              throw new RuntimeException(errorMessage);
+            }
+          });
+
+      try {
+        // Only set the config reference as this value came from the cache
+        config = bq.take();
```
aarsilv, line 78: Oh cool! So this will wait until something has been added to the queue? Nifty.

I wonder if we should have timeouts for this stuff, settable in Configuration but with some reasonable defaults.

src/main/java/cloud/eppo/ConfigurationWrangler.java
```diff
@@ -0,0 +1,241 @@
...
+    setConfig(requestor.load(config));
+  }
+
+  // Load Async Method
+  // Start an async load via the configStore (which can be subclassed by libraries using this
+  // package to load from disk/cache/etc)
+  // Also start an async http fetch (this is required for Android to have network calls off the main
+  // thread, which is a good idea anyway).
+  //
+  // The precedence strategy is first-to-succeed, which is to say, the first loader to succeed gets
```
aarsilv, line 93: 🐎

src/main/java/cloud/eppo/ConfigurationWrangler.java
```diff
@@ -0,0 +1,241 @@
...
+          public void onFailure(String errorMessage) {
+            lock.lock();
+            try {
+              cacheLoadInProgress.set(false);
+              log.debug("Did not initialize from cache");
+              // If cache loading failed, and fetching failed, fire off the failure callback if not
+              // yet done so
+              // Otherwise, if fetching has not failed yet, defer to it for firing off callbacks
+              if (callback != null
+                  && fetchErrorMessage.get() != null
```
aarsilv, line 150: I don't think we need to make sure `fetchErrorMessage` is non-null here. We should only be checking if a callback is at play in case something else goes wonky

src/main/java/cloud/eppo/EppoHttpClient.java
aarsilv: I love how something added in 2011 is "modern" 😛

src/main/java/cloud/eppo/IConfigurationStore.java
```diff
@@ -0,0 +1,12 @@
+package cloud.eppo;
+
+import cloud.eppo.api.Configuration;
+import org.jetbrains.annotations.NotNull;
+
+public interface IConfigurationStore {
```
aarsilv, line 6: This is where `isExpired()` would be added if we want to go that route

src/main/java/cloud/eppo/api/Configuration.java
```diff
@@ -115,10 +122,15 @@ public boolean isConfigObfuscated() {
     return isConfigObfuscated;
   }
 
+  @SuppressWarnings("unused")
+  public byte[] getFlagConfigJson() {
```
aarsilv, line 126: I wonder if we should call this `getFlagCOnfigJsonBytes()`

aarsilv (commented):
Looking good! Left MANY musings that I'd love for you to consider.
Also, as you pointed out, we'll want tests of the various race conditions. You hopefully can steal them from the current Android SDK.

Lastly--if not too crazy--consider mirroring the functionality of the JavaScript client with regards to its configurable use of expired stores and update strategies. However this could also be done later along with iOS.

src/main/java/cloud/eppo/ConfigurationRequestor.java
typotter: 📈

src/main/java/cloud/eppo/BaseEppoClient.java
```diff
@@ -24,9 +25,14 @@ public class BaseEppoClient {
...
   protected static final String DEFAULT_HOST = "https://fscdn.eppo.cloud";
-  protected final ConfigurationRequestor requestor;
 
-  private final ConfigurationStore configurationStore;
+  /**
+   * The Configuration Wrangler handles the complexities of loading configuration from multiple
+   * sources and saving to a persistent store. It contains the source of truth reference to the
+   * **Configuration** (UFC definitions and Bandit Parameters).
+   */
+  protected final ConfigurationWrangler configurationWrangler;
```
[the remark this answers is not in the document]
typotter: not at all. I went with `Manager` because this class does more than just load configs (it has to manager, or wrangle the async config loading and apply an update strategy). tbh, I've just grown a little bored of `Manager` so look for ways to spice it up. maybe "Coordinator"...

src/main/java/cloud/eppo/BaseEppoClient.java
```diff
@@ -468,4 +512,6 @@ private <T> T throwIfNotGraceful(Exception e, T defaultValue) {
   public void setIsGracefulFailureMode(boolean isGracefulFailureMode) {
     this.isGracefulMode = isGracefulFailureMode;
   }
+
+  public interface InitializationCallback extends EppoCallback<Void> {}
```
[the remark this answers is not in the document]
typotter: Wouldn't hurt. This file is already pretty large.

src/main/java/cloud/eppo/ConfigurationRequestor.java
```diff
@@ -1,55 +1,101 @@
...
 
-import java.io.IOException;
-import okhttp3.Response;
+import cloud.eppo.api.Configuration;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 // TODO: handle bandit stuff
-public class ConfigurationRequestor {
+class ConfigurationRequestor {
```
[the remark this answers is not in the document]
typotter: I meant to ask what stuff was still needed here.

src/main/java/cloud/eppo/ConfigurationWrangler.java
typotter: That's right. This seemed like the cleanest way to turn the callbacks into sync. Java `Future`s have a lot of this built-in, but that would require the SDK be bumped up to 24 (from 21) in the Android SDK :( There is a way to only wait a specified amount of time before giving up. We'll have to add another parameter to the `BaseEppoClient` constructor to plumb through the timeout, but that's cleaned up by the `Builder`s in the java-server- and android SDKs

src/main/java/cloud/eppo/ConfigurationWrangler.java
typotter: The `fetchErrorMessage` + callbackCalled.getAndSet are a sufficient proxy for something like a `fetchCallInProgress` boolean and are variables that we need to track anyway. An additional atomicBoolean would be more clear, if a little redundant.
Happy to change it, but I thought your original comments and code were sufficient here.

src/main/java/cloud/eppo/api/Configuration.java
typotter: There is only one `getFlagConfigJson` method (so nothing to disambiguate) and the type is right there so I didn't think adding `Bytes` to the name would really add much. OTOH, extra description is rarely hurtful.

src/main/java/cloud/eppo/IConfigurationStore.java
typotter: Let's iterate on this PR to get to a solid footing for migrating Android, then I'll create a PR to `isExpired` logic into this branch (or into main).

src/main/java/cloud/eppo/ConfigurationWrangler.java
typotter: Agreed this is a good time, but let's separate that from this PR for now and I'll send a follow up right behind it.

src/main/java/cloud/eppo/BaseEppoClient.java
```diff
@@ -14,6 +14,7 @@
 import com.fasterxml.jackson.databind.ObjectMapper;
 import java.util.HashMap;
 import java.util.Map;
+import org.jetbrains.annotations.Nullable;
```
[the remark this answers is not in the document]
typotter: actually, I think this is just from me dipping back and forth from android where `androidx.annotations` are used.

src/main/java/cloud/eppo/BaseEppoClient.java
```diff
@@ -68,7 +76,10 @@ protected BaseEppoClient(
       BanditLogger banditLogger,
       boolean isGracefulMode,
       boolean expectObfuscatedConfig,
-      Configuration initialConfiguration) {
+      EppoHttpClient httpClient,
```
felipecsl, line 79: I know this is something we haven't done at Eppo, but I'd like us to start using `@Nullable` and `@NotNull` annotations to indicate which parameters are optional and which ones arent, to help with readability and also IDEs like IntelliJ can use that information to assist with finding bugs. Given `httpClient` is optional, can you please add a `@Nullable` annotation here? Same for `apiKey`, `configStoreOverride` and any other optional params here. It's fine to do incrementally/handle in a separate PR if you prefer

src/main/java/cloud/eppo/ConfigurationManager.java
```diff
@@ -0,0 +1,244 @@
...
+import cloud.eppo.api.Configuration;
+import java.util.concurrent.BlockingQueue;
+import java.util.concurrent.LinkedBlockingQueue;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.concurrent.atomic.AtomicReference;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class ConfigurationManager {
```
felipecsl, line 12: add a 1-2 paragraph comment describing the purpose of this class. also seems like this shouldn't be public (i don't see any public methods)

src/main/java/cloud/eppo/ConfigurationManager.java
```diff
@@ -0,0 +1,244 @@
...
+      ConfigurationRequestor requestor,
+      IConfigurationStore persistentSource,
+      Configuration initialConfiguration) {
+    this(requestor, persistentSource, initialConfiguration, 30);
+  }
+
+  ConfigurationManager(
+      ConfigurationRequestor requestor,
+      IConfigurationStore persistentSource,
+      Configuration initialConfiguration,
```
felipecsl, line 32: `@Nullable`

src/main/java/cloud/eppo/EppoHttpClient.java
felipecsl: there's a tradeoff here in which we can't stream the data and have to load it all in memory. seems fine given it doesn't seem like it's a massive amount of data, but something to keep in mind.

felipecsl: Alternatively you could return an `InputStream` or equivalent, as long as you ensure the caller closes it properly

felipecsl (commented):
No blocking comments 👍🏽

typotter:
closing in favour of #41

PR #41, opened 2024-09-24
feat: Async configuration handling using CompletableFutures and other config refactoring

typotter:
## Motivation and Context
- Android is migrating to use the common SDK and has some additional requirements around asynchronous fetching, asynchronous initial configuration and resolving race conditions from config sources (cache vs remote API).

- The race condition between a cache load and a remote fetch only happens during the initialization of the `BaseEppoClient`. Specifically, when the client is created with an initial configuration and then the async `loadConfiguration` method is called. In this case, we don't want the cache load to clobber the fetched configuration as the fetched flags are always the latest. On subsequent calls to load (or loadAsync), we already have a reference to a `Configuration` in memory, so a cache-load attempt is not needed.

- The `IConfigurationStore` interface from the javascript common library is a robust interface capable of implementing multiple different config loading and caching strategies. Similar flexibility in the java common sdk is the goal; this change is a step towards it that will allow the additional complexities to be layered on.

- `CompletableFuture` gives a nice API to track done-ness without the need for atomics.

## Overview of changes
- Make `initialFlagConfig` parameter a `CompletableFuture<Configuration>`. This facilitates initializing the client with a config string as well as initial loads from cache.
- `CompletableFuture`s used in place of callback in `EppoHttpClient`, `ConfigurationRequestor`, etc.
- `IconfigurationStore` interface for extending cache and config storage strategies.
- moved `Configuration` to `api` package
- Async `fetchAndSaveFromRemote` method
- Initial Configuration setting is pushed to the `ConfigurationRequestor` so that it can be properly raced with async fetch calls

### DIFF
--- a/src/main/java/cloud/eppo/BaseEppoClient.java
+++ b/src/main/java/cloud/eppo/BaseEppoClient.java
@@ -127,9 +127,6 @@ public class BaseEppoClient {
     // Save SDK name and version to include in logger metadata
     this.sdkName = sdkName;
     this.sdkVersion = sdkVersion;
-
-    // TODO: caching initialization (such as setting an API-key-specific prefix
-    //       will probably involve passing in configurationStore to the constructor
   }
 
   private EppoHttpClient buildHttpClient(
--- a/src/main/java/cloud/eppo/BaseEppoClient.java
+++ b/src/main/java/cloud/eppo/BaseEppoClient.java
@@ -26,7 +26,7 @@ public class BaseEppoClient {
   protected static final String DEFAULT_HOST = "https://fscdn.eppo.cloud";
   protected final ConfigurationRequestor requestor;
 
-  private final ConfigurationStore configurationStore;
+  private final IConfigurationStore configurationStore;
   private final AssignmentLogger assignmentLogger;
   private final BanditLogger banditLogger;
   private final String sdkName;
--- a/src/main/java/cloud/eppo/BaseEppoClient.java
+++ b/src/main/java/cloud/eppo/BaseEppoClient.java
@@ -14,6 +14,7 @@ import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import java.util.HashMap;
 import java.util.Map;
+import java.util.concurrent.CompletableFuture;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -111,10 +112,12 @@ public class BaseEppoClient {
   }
 
   protected void loadConfiguration() {
-    requestor.load();
+    requestor.fetchAndSaveFromRemote();
   }
 
-  // TODO: async way to refresh for android
+  protected CompletableFuture<Void> loadConfigurationAsync() {
+    return requestor.fetchAndSaveFromRemoteAsync();
+  }
 
   protected EppoValue getTypedAssignment(
       String flagKey,
--- a/src/main/java/cloud/eppo/BaseEppoClient.java
+++ b/src/main/java/cloud/eppo/BaseEppoClient.java
@@ -94,7 +94,7 @@ public class BaseEppoClient {
       IConfigurationStore configurationStore,
       boolean isGracefulMode,
       boolean expectObfuscatedConfig,
-      Configuration initialConfiguration) {
+      CompletableFuture<Configuration> initialConfiguration) {
 
     if (apiKey == null) {
       throw new IllegalArgumentException("Unable to initialize Eppo SDK due to missing API key");
@@ -109,7 +109,7 @@ public class BaseEppoClient {
 
     EppoHttpClient httpClient = buildHttpClient(host, apiKey, sdkName, sdkVersion);
     if (configurationStore == null) {
-      this.configurationStore = new ConfigurationStore(initialConfiguration);
+      this.configurationStore = new ConfigurationStore();
     } else {
       this.configurationStore = configurationStore;
     }
@@ -117,6 +117,10 @@ public class BaseEppoClient {
     // For now, the configuration is only obfuscated for Android clients
     requestor =
         new ConfigurationRequestor(this.configurationStore, httpClient, expectObfuscatedConfig);
+    if (initialConfiguration != null) {
+      requestor.setInitialConfiguration(initialConfiguration);
+    }
+
     this.assignmentLogger = assignmentLogger;
     this.banditLogger = banditLogger;
     this.isGracefulMode = isGracefulMode;
--- a/src/main/java/cloud/eppo/BaseEppoClient.java
+++ b/src/main/java/cloud/eppo/BaseEppoClient.java
@@ -15,6 +15,8 @@ import com.fasterxml.jackson.databind.ObjectMapper;
 import java.util.HashMap;
 import java.util.Map;
 import java.util.concurrent.CompletableFuture;
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -40,61 +42,17 @@ public class BaseEppoClient {
   private static EppoHttpClient httpClientOverride = null;
 
   protected BaseEppoClient(
-      String apiKey,
-      String sdkName,
-      String sdkVersion,
-      String host,
-      AssignmentLogger assignmentLogger,
-      BanditLogger banditLogger,
-      boolean isGracefulMode,
-      boolean expectObfuscatedConfig) {
-    this(
-        apiKey,
-        sdkName,
-        sdkVersion,
-        host,
-        assignmentLogger,
-        banditLogger,
-        null,
-        isGracefulMode,
-        expectObfuscatedConfig,
-        null);
-  }
-
-  protected BaseEppoClient(
-      String apiKey,
-      String sdkName,
-      String sdkVersion,
-      String host,
-      AssignmentLogger assignmentLogger,
-      BanditLogger banditLogger,
-      IConfigurationStore configurationStore,
-      boolean isGracefulMode,
-      boolean expectObfuscatedConfig) {
-    this(
-        apiKey,
-        sdkName,
-        sdkVersion,
-        host,
-        assignmentLogger,
-        banditLogger,
-        configurationStore,
-        isGracefulMode,
-        expectObfuscatedConfig,
-        null);
-  }
-
-  protected BaseEppoClient(
-      String apiKey,
-      String sdkName,
-      String sdkVersion,
-      String host,
-      AssignmentLogger assignmentLogger,
-      BanditLogger banditLogger,
-      IConfigurationStore configurationStore,
+      @NotNull String apiKey,
+      @NotNull String sdkName,
+      @NotNull String sdkVersion,
+      @Nullable String host,
+      @Nullable AssignmentLogger assignmentLogger,
+      @Nullable BanditLogger banditLogger,
+      @Nullable IConfigurationStore configurationStore,
       boolean isGracefulMode,
       boolean expectObfuscatedConfig,
-      CompletableFuture<Configuration> initialConfiguration) {
+      boolean supportBandits,
+      @Nullable CompletableFuture<Configuration> initialConfiguration) {
 
     if (apiKey == null) {
       throw new IllegalArgumentException("Unable to initialize Eppo SDK due to missing API key");
@@ -108,15 +66,13 @@ public class BaseEppoClient {
     }
 
     EppoHttpClient httpClient = buildHttpClient(host, apiKey, sdkName, sdkVersion);
-    if (configurationStore == null) {
-      this.configurationStore = new ConfigurationStore();
-    } else {
-      this.configurationStore = configurationStore;
-    }
+    this.configurationStore =
+        configurationStore != null ? configurationStore : new ConfigurationStore();
 
     // For now, the configuration is only obfuscated for Android clients
     requestor =
-        new ConfigurationRequestor(this.configurationStore, httpClient, expectObfuscatedConfig);
+        new ConfigurationRequestor(
+            this.configurationStore, httpClient, expectObfuscatedConfig, supportBandits);
     if (initialConfiguration != null) {
       requestor.setInitialConfiguration(initialConfiguration);
     }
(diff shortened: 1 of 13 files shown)


src/main/java/cloud/eppo/EppoHttpClient.java
```diff
@@ -41,34 +42,23 @@ private static OkHttpClient buildOkHttpClient() {
...
-  // TODO: use this for Java, callback for Android; clean as needed
-  public Response get(String path) {
-    HttpUrl httpUrl =
-        HttpUrl.parse(baseUrl + path)
-            .newBuilder()
-            .addQueryParameter("apiKey", apiKey)
-            .addQueryParameter("sdkName", sdkName)
-            .addQueryParameter("sdkVersion", sdkVersion)
-            .build();
+  public byte[] get(String path) {
```
typotter, line 45: wrt Streaming the data, unfortunately, we do need to load the entire contents into memory in order to parse it due to the nature of serialized JSON.

src/main/java/cloud/eppo/EppoHttpClient.java
```diff
@@ -41,34 +42,23 @@ private static OkHttpClient buildOkHttpClient() {
...
+    Request request = buildRequest(path);
 
-    Request request = new Request.Builder().url(httpUrl).build();
-    try {
-      return client.newCall(request).execute();
+    try (Response response = client.newCall(request).execute()) {
```
typotter, line 48: Alternatively, we could call and wait for the async method to complete.

src/main/java/cloud/eppo/BaseEppoClient.java
```diff
@@ -14,6 +14,7 @@
 import com.fasterxml.jackson.databind.ObjectMapper;
 import java.util.HashMap;
 import java.util.Map;
+import java.util.concurrent.CompletableFuture;
```
aarsilv, line 17: 🔥

src/main/java/cloud/eppo/BaseEppoClient.java
```diff
@@ -66,9 +68,33 @@ protected BaseEppoClient(
...
+        host,
+        assignmentLogger,
+        banditLogger,
+        configurationStore,
+        isGracefulMode,
+        expectObfuscatedConfig,
+        null);
+  }
+
+  protected BaseEppoClient(
```
aarsilv, line 87: given this constructor is protected and will only be used by upstream SDK builders (and tests here), I don't think we need to make `initialConfiguration` optional and can just require `null` or a `CompletableFuture` to be passed in to a single constructor.

src/main/java/cloud/eppo/BaseEppoClient.java
```diff
@@ -82,19 +108,25 @@ protected BaseEppoClient(
     }
 
     EppoHttpClient httpClient = buildHttpClient(host, apiKey, sdkName, sdkVersion);
-    this.configurationStore = new ConfigurationStore(initialConfiguration);
+    if (configurationStore == null) {
+      this.configurationStore = new ConfigurationStore();
+    } else {
+      this.configurationStore = configurationStore;
+    }
```
aarsilv, line 115: alternatively, there is the ternary operator 😛 
```java
this.configurationStore = configurationStore != null ? configurationStore : new ConfigurationStore();
```

src/main/java/cloud/eppo/BaseEppoClient.java
```diff
@@ -82,19 +108,25 @@ protected BaseEppoClient(
...
 
     // For now, the configuration is only obfuscated for Android clients
-    requestor = new ConfigurationRequestor(configurationStore, httpClient, expectObfuscatedConfig);
+    requestor =
+        new ConfigurationRequestor(this.configurationStore, httpClient, expectObfuscatedConfig);
+    if (initialConfiguration != null) {
+      requestor.setInitialConfiguration(initialConfiguration);
```
aarsilv, line 121: 💪

src/main/java/cloud/eppo/BaseEppoClient.java
```diff
@@ -111,10 +143,12 @@ private EppoHttpClient buildHttpClient(
   }
 
   protected void loadConfiguration() {
-    requestor.load();
+    requestor.fetchAndSaveFromRemote();
```
aarsilv, line 146: 👌

src/main/java/cloud/eppo/BaseEppoClient.java
```diff
@@ -111,10 +143,12 @@ private EppoHttpClient buildHttpClient(
...
   }
 
-  // TODO: async way to refresh for android
+  protected CompletableFuture<Void> loadConfigurationAsync() {
+    return requestor.fetchAndSaveFromRemoteAsync();
+  }
```
aarsilv, line 151: 👍

src/main/java/cloud/eppo/ConfigurationRequestor.java
```diff
@@ -1,55 +1,148 @@
...
   public ConfigurationRequestor(
-      ConfigurationStore configurationStore,
-      EppoHttpClient client,
+      @NotNull IConfigurationStore configurationStore,
+      @NotNull EppoHttpClient client,
       boolean expectObfuscatedConfig) {
     this.configurationStore = configurationStore;
     this.client = client;
     this.expectObfuscatedConfig = expectObfuscatedConfig;
+    this.supportBandits = true;
```
aarsilv, line 43: Same here, I think we can just deal with a singular constructor. The "developer experience" win is limited to our SDK code and doesn't seem worth the added overhead of multiple constructors.

src/main/java/cloud/eppo/ConfigurationRequestor.java
```diff
@@ -1,55 +1,148 @@
...
+        configurationFuture.thenAccept(
+            (config) -> {
+              synchronized (this) {
+                // Don't clobber an actual fetch.
+                if (config == null) {
+                  log.debug("Initial configuration future returned null");
+                } else if (remoteFetchFuture != null && remoteFetchFuture.isDone()) {
+                  log.debug("Fetch beat the initial config; not clobbering");
+                } else {
+                  configurationStore.saveConfiguration(config);
```
aarsilv, line 76: do we want a `log.debug()` above this line for completeness?
```java
log.debug("saving initial configuration");
```

src/main/java/cloud/eppo/ConfigurationRequestor.java
```diff
@@ -1,55 +1,148 @@
...
+            .getAsync(FLAG_CONFIG_PATH)
+            .thenApply(
+                flagConfigJsonBytes -> {
+                  synchronized (this) {
+                    Configuration.Builder configBuilder =
+                        new Configuration.Builder(flagConfigJsonBytes, expectObfuscatedConfig)
+                            .banditParametersFromConfig(
+                                lastConfig); // possibly reuse last bandit models loaded.
+
+                    if (supportBandits && configBuilder.requiresBanditModels()) {
```
aarsilv, line 126: I wonder if we should rename `requiresBanditModels()` to `requiresUpdatedBanditModels()`, which (I think) is more explicit about what it is considering and makes the above code comment make more sense.

src/main/java/cloud/eppo/ConfigurationRequestor.java
```diff
@@ -1,55 +1,148 @@
...
+                      if (banditParametersJsonBytes != null) {
+                        configBuilder.banditParameters(banditParametersJsonBytes);
+                      }
+                    }
+                    return configBuilder.build();
+                  }
+                })
+            .thenApply(
+                configuration -> {
+                  configurationStore.saveConfiguration(configuration);
```
aarsilv, line 143: Should this be in a `synchronized` block too?

src/main/java/cloud/eppo/EppoHttpClient.java
```diff
@@ -78,16 +68,19 @@ public void onResponse(Call call, Response response) {
...
                   } catch (IOException ex) {
-                    callback.onFailure("Failed to read response from URL " + httpUrl);
+                    future.completeExceptionally(
+                        new RuntimeException(
+                            "Failed to read response from URL {}" + request.url(), ex));
                   }
                 } else {
                   if (response.code() == HttpURLConnection.HTTP_FORBIDDEN) {
-                    callback.onFailure("Invalid API key");
+                    future.completeExceptionally(new RuntimeException("Invalid API key"));
```
aarsilv, line 79: I know it was like this when you found it, but I wonder why do these additional checks with the asynchronous method but not the synchronous one.

Without looking too closely, it seems this code duplication shouldn't be needed and we could share most code with the synchronous `get()`, leveraging either calling it in a wrapped way, or some shared private base method.

src/test/java/cloud/eppo/BaseEppoClientTest.java
```diff
@@ -258,6 +259,28 @@ public void testWithInitialConfiguration() {
     }
   }
 
+  @Test
+  public void testWithInitialConfigurationFuture() {
```
aarsilv, line 263: 🙌

src/test/java/cloud/eppo/BaseEppoClientTest.java
```diff
@@ -311,5 +334,32 @@ public void testAssignmentLogErrorNonFatal() {
     verify(mockAssignmentLogger, times(1)).logAssignment(assignmentLogCaptor.capture());
   }
 
+  @Test
+  public void testConfigurationStoreOverride() throws IOException {
```
aarsilv, line 338: What is this test testing exactly? That if the configuration changes after initialization we get updated assignments?

src/test/java/cloud/eppo/ConfigurationRequestorTest.java
```diff
@@ -0,0 +1,79 @@
...
+    assertNull(configStore.getConfiguration());
+
+    futureConfig.complete(new Configuration.Builder(flagConfig, false).build());
+
+    assertNotNull(configStore.getConfiguration());
+    assertNotNull(configStore.getConfiguration().getFlag("numeric_flag"));
+  }
+
+  @Test
+  public void testInitialConfigurationDoesntClobberFetch() throws IOException {
```
aarsilv, line 44: 🙌

src/test/java/cloud/eppo/ConfigurationRequestorTest.java
```diff
@@ -0,0 +1,79 @@
...
+
+    // The initial config contains only one flag keyed `numeric_flag`. The fetch response has only
+    // one flag keyed
+    // `boolean_flag`. We make sure to complete the fetch future first to verify the cache load does
+    // not overwrite it.
+    CompletableFuture<Void> handle = requestor.fetchAndSaveFromRemoteAsync();
+
+    // Resolve the fetch and then the initialConfig
+    configFetchFuture.complete(fetchedFlagConfig.getBytes(StandardCharsets.UTF_8));
+    initialConfigFuture.complete(new Configuration.Builder(flagConfig, false).build());
```
aarsilv, line 72: 👍 so this is what we expect to be ignored

aarsilv (approved):
I like the iteration and the interface for upstream APIs! 💪  Also love the tests 🙌

src/test/java/cloud/eppo/BaseEppoClientTest.java
typotter: It's testing that the client is actually using the `configurationStoreOverride argument. Not really necessary, I suppose. When the ConfigStore has more capabilities, those tests will rely on being able to pass in a ConfigStore (the android SDK will also prove this out).

src/main/java/cloud/eppo/BaseEppoClient.java
```diff
@@ -82,39 +66,39 @@ protected BaseEppoClient(
...
     this.sdkName = sdkName;
     this.sdkVersion = sdkVersion;
-
-    // TODO: caching initialization (such as setting an API-key-specific prefix
-    //       will probably involve passing in configurationStore to the constructor
   }
 
   private EppoHttpClient buildHttpClient(
       String host, String apiKey, String sdkName, String sdkVersion) {
-    EppoHttpClient httpClient;
```
typotter, line 102: 😉

typotter (commented):
Thanks @aarsilv

src/main/java/cloud/eppo/EppoHttpClient.java
felipecsl: can you expand on "due to the nature of serialized JSON"? Just curious about where the issue is

src/main/java/cloud/eppo/BaseEppoClient.java
```diff
@@ -39,36 +42,17 @@ public class BaseEppoClient {
...
-      String host,
-      AssignmentLogger assignmentLogger,
-      BanditLogger banditLogger,
+      @NotNull String apiKey,
+      @NotNull String sdkName,
+      @NotNull String sdkVersion,
+      @Nullable String host,
+      @Nullable AssignmentLogger assignmentLogger,
+      @Nullable BanditLogger banditLogger,
+      @Nullable IConfigurationStore configurationStore,
```
felipecsl, line 51: 👌🏽

src/test/java/cloud/eppo/BaseEppoClientTest.java
felipecsl: nice test

src/test/java/cloud/eppo/BaseEppoClientTest.java
```diff
@@ -258,6 +262,28 @@ public void testWithInitialConfiguration() {
...
+
+      // Now, complete the initial config future and check the value.
+      futureConfig.complete(Configuration.builder(flagConfig, false).build());
+
+      result = eppoClient.getDoubleAssignment("numeric_flag", "dummy subject", 0);
+      assertEquals(5, result);
+
+    } catch (IOException e) {
+      throw new RuntimeException(e);
+    }
```
felipecsl, line 284: nit: you could just add `throws IOException` to the method signature

felipecsl (approved):
💯

src/main/java/cloud/eppo/EppoHttpClient.java
typotter: I stand corrected, thank you @felipecsl for pushing and teaching :) Really do appreciate it.
Now that we've established we can in fact use a byte stream, should we do it in this PR, a followup, or keep it on the backburner?

felipecsl: i'm okay keeping in the backburner. not a blocker from my perspective. thanks for not dropping the thread :)

PR #43, opened 2024-10-02
feat/fix: Configuration enhancements

typotter:
## Motivation and Context
Make saving configurations in downstream APIs easier

## Overview of changes
- "grafting" `forServer` field onto the flag config response JSON so that it can be stored with the FCR in a persistent store

### DIFF
--- a/src/main/java/cloud/eppo/api/Configuration.java
+++ b/src/main/java/cloud/eppo/api/Configuration.java
@@ -6,7 +6,6 @@ import cloud.eppo.ufc.dto.*;
 import cloud.eppo.ufc.dto.adapters.EppoModule;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import java.io.IOException;
-import java.nio.charset.StandardCharsets;
 import java.util.Collections;
 import java.util.Map;
 import java.util.Set;
@@ -116,8 +115,12 @@ public class Configuration {
     return isConfigObfuscated;
   }
 
-  public String getFlagConfigJsonString() {
-    return new String(flagConfigJson, StandardCharsets.UTF_8);
+  public byte[] serializeFlagConfigToBytes() {
+    return flagConfigJson;
+  }
+
+  public byte[] serializeBanditParamsToBytes() {
+    return banditParamsJson;
   }
 
   /**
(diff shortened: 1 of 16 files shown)


src/main/java/cloud/eppo/api/Configuration.java
```diff
@@ -150,8 +154,11 @@ public Builder(byte[] flagJson, boolean isConfigObfuscated) {
       this.isConfigObfuscated = isConfigObfuscated;
 
       if (flagJson == null || flagJson.length == 0) {
-        throw new RuntimeException(
-            "Null or empty configuration string. Call `Configuration.Empty()` instead");
+        log.warn("Null or empty configuration string. Call `Configuration.Empty()` instead");
```
aarsilv, line 157: Ah yeah having this non-fatal is better dev ex.

src/main/java/cloud/eppo/api/Configuration.java
```diff
@@ -230,4 +241,45 @@ public Configuration build() {
...
   }
+
+  /**
+   * Contains the base data required to hydrate a `Configuration` object
+   *
+   * <p>Using `Configuration.writeToSteam` and `Configuration.readFromStream` allows us to avoid
+   * implementing JSON serialization for the `Configuration` class or making all the DTOs `implement
+   * Serializable`
+   */
+  private static class ConfigurationPacket implements Serializable {
```
aarsilv, line 252: oh I like this name

src/main/java/cloud/eppo/api/Configuration.java
```diff
@@ -230,4 +241,45 @@ public Configuration build() {
...
+      this.isConfigObfuscated = isConfigObfuscated;
+    }
+  }
+
+  private ConfigurationPacket getConfigurationPacket() {
+    return new ConfigurationPacket(flagConfigJson, banditParamsJson, isConfigObfuscated);
+  }
+
+  public void writeToStream(OutputStream out) throws IOException {
+    ObjectOutputStream oos = new ObjectOutputStream(out);
```
aarsilv, line 270: ❗ didn't know about this

However is this the planned use for caching in the filesystem? I think for that we should just save off the original received JSON

aarsilv (approved):

src/main/java/cloud/eppo/api/Configuration.java
typotter: refactored everything to save the JSON bytes directly to disk. Requires "grafting" the `forServer` field onto the `FlagConfigResponse` json which in turn, requires a deserialization+reserialization before the JSON string is written to disk. This, of course, will be unnecessary when `forServer` is added to the FCR

PR #45, opened 2024-10-03
chore: Change 'forServer' boolean to format enum

typotter:


aarsilv (approved):
Relevant: https://github.com/Eppo-exp/eppo/pull/10896

src/main/java/cloud/eppo/ufc/dto/FlagConfigResponse.java
```diff
@@ -34,7 +34,12 @@ public Map<String, BanditReference> getBanditReferences() {
...
 
-  public boolean isForServer() {
-    return this.forServer;
+  public Format getFormat() {
+    return format;
+  }
+
+  public enum Format {
+    SERVER,
+    CLIENT
```
rasendubi, line 43: Is this replacing obfuscation flag? If yes, it seems to be a wrong dichotomy because clients may work with non-obfuscated configs, and servers may work with obfuscated ones.

typotter: yes, this replaces the flag. 
see https://github.com/Eppo-exp/eppo/pull/10896

PR #50, opened 2024-10-11
feat: assignment cache

typotter:
Eppo Internal
:tickets: towards FF-2280

## Motivation and Context
Deduplication of Assignment logs in the Java-based SDKs

## Description of Changes
- new `IAssignmentCache` interface for storing assignments. It is up to the implementation to decide cache size, age, etc.
- `MapAssignmentCache` basic implementation to use in tests

## How has this been tested
- Test cases in `BaseEppoClientTest` to demonstrate duplicated logging calls and test deduplication.

## What is expected of reviewers
- 👀


src/main/java/cloud/eppo/BaseEppoClient.java
```diff
@@ -176,10 +180,18 @@ protected EppoValue getTypedAssignment(
...
-        log.warn("Error logging assignment: {}", e.getMessage(), e);
+
+      // Deduping assignment logging is possible by providing an `IAssignmentCache`
+      String assignmentCacheKey = assignment.getIdentifier();
+      if (assignmentCache == null || !assignmentCache.containsKey(assignmentCacheKey)) {
+        try {
+          assignmentLogger.logAssignment(assignment);
+          if (assignmentCache != null) {
+            assignmentCache.put(assignmentCacheKey, assignment);
+          }
```
aarsilv, line 191: 👍

src/main/java/cloud/eppo/api/IAssignmentCache.java
```diff
@@ -0,0 +1,15 @@
+package cloud.eppo.api;
+
+import cloud.eppo.logging.Assignment;
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
+public interface IAssignmentCache {
+  void put(@NotNull String key, @NotNull Assignment assignment);
+
+  @Nullable Assignment get(@NotNull String key);
```
aarsilv, line 10: I wonder if we should explicitly have it take a `flagKey`, `subjectKey`, `allocationKey`, and `variationKey` vs. a single string.

src/main/java/cloud/eppo/logging/Assignment.java
```diff
@@ -33,6 +38,12 @@ public Assignment(
     this.subjectAttributes = subjectAttributes;
     this.extraLogging = extraLogging;
     this.metaData = metaData;
+
+    variationKeyString = subject + sep + featureFlag + sep + variation;
```
aarsilv, line 42: In our JavaScript SDK we also include the allocation, although I'm not sure why 😅

src/test/java/cloud/eppo/BaseEppoClientTest.java
```diff
@@ -317,6 +343,42 @@ public void testAssignmentEventCorrectlyCreated() {
     assertEquals(expectedMeta, capturedAssignment.getMetaData());
   }
 
+  @Test
+  public void testAssignmentNotDedupedWithoutCache() {
```
aarsilv, line 347: 🙌

src/test/java/cloud/eppo/BaseEppoClientTest.java
```diff
@@ -317,6 +343,42 @@ public void testAssignmentEventCorrectlyCreated() {
...
+    subjectAttributes.put("employer", EppoValue.valueOf("Eppo"));
+
+    // Get the assignment twice
+    eppoClient.getDoubleAssignment("numeric_flag", "alice", subjectAttributes, 0.0);
+    eppoClient.getDoubleAssignment("numeric_flag", "alice", subjectAttributes, 0.0);
+
+    ArgumentCaptor<Assignment> assignmentLogCaptor = ArgumentCaptor.forClass(Assignment.class);
+
+    // `logAssignment` should be called only once.
+    verify(mockAssignmentLogger, times(1)).logAssignment(assignmentLogCaptor.capture());
```
aarsilv, line 379: I'd extend this test to then assign a different number, and then the original. Because of the flip-flop, each assignment should be logged as new.

This capture a nuance in that the _key_ should be subject key and flag key, but the value should include the variation. That way new variations evict old ones.

aarsilv (changes requested):
This is a good start, but the cache is not as nuanced as we need. Consider following the same pattern in JavaScript client where we have an `AbstractAssignmentCache` ([source](https://github.com/Eppo-exp/js-client-sdk-common/blob/main/src/cache/abstract-assignment-cache.ts)). It would handle making the keys just the flag and subject, and the values allocation and variation keys. (Note variation key is different than variation value--presently just for JSON-typed values but perhaps one day all values).

Then `MapAssignmentCache` and others would extend that.

src/test/java/cloud/eppo/BaseEppoClientTest.java
typotter: added. thanks for the precision.

src/test/java/cloud/eppo/BaseEppoClientTest.java
```diff
@@ -317,6 +346,67 @@ public void testAssignmentEventCorrectlyCreated() {
...
+    // Get the assignment
+    int newAssignment =
+        eppoClient.getIntegerAssignment("numeric-one-of", "alice", subjectAttributes, 0);
+    assertEquals(1, newAssignment);
+
+    // Verify a new log call
+    verify(mockAssignmentLogger, times(2)).logAssignment(any(Assignment.class));
+
+    // Change back to the original variation to ensure it is not still cached after the previous
+    // value evicted it.
```
aarsilv, line 398: 🔥

aarsilv (commented):
This is getting closer--thanks for iterating! Appologies that you're chasing a moving target here, but I just caught up with @schmit, and he agreed we want to log all "distinct" bandit actions. 

So we'll want, by default, expiring caches for bandit _actions_. Bandit _variations_ can use the normal assignment cache. The expiration time should be short, such as 10 minutes.

I know how frustrating this is and appreciate you dealing with these on-the-fly context and requirements changes 🙏

src/main/java/cloud/eppo/cache/NonExpiringInMemoryAssignmentCache.java
```diff
@@ -0,0 +1,16 @@
...
+import cloud.eppo.api.AbstractAssignmentCache;
+import java.util.concurrent.ConcurrentHashMap;
+
+/**
+ * A cache that never expires.
+ *
+ * <p>The primary use case is for client-side SDKs, where the cache is only used for a single user
+ * and won't grow out of control.
+ */
+public class NonExpiringInMemoryAssignmentCache extends AbstractAssignmentCache {
```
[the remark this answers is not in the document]
typotter: The apache commons `LRUMap` makes for a really straightforward implementation using the `AbstractAssignmentCache`.

src/main/java/cloud/eppo/cache/AssignmentCacheKey.java
```diff
@@ -0,0 +1,45 @@
...
+    return subjectKey;
+  }
+
+  public String getFlagKey() {
+    return flagKey;
+  }
+
+  @Override
+  public String toString() {
+    return subjectKey + "-" + flagKey;
```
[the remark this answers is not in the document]
typotter: agreed.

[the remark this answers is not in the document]
typotter: changed to semicolon separator.

src/main/java/cloud/eppo/cache/BanditCacheValue.java
```diff
@@ -0,0 +1,31 @@
+package cloud.eppo.cache;
+
+import java.util.Objects;
+
+public class BanditCacheValue implements AssignmentCacheValue {
```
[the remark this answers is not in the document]
typotter: leaving the cache in place with an expiring cache implementation available to downstream SDKs.

src/main/java/cloud/eppo/cache/ExpiringInMemoryAssignmentCache.java
```diff
@@ -0,0 +1,18 @@
...
+import java.util.Map;
+import java.util.concurrent.TimeUnit;
+import org.apache.commons.collections4.map.PassiveExpiringMap;
+
+public class ExpiringInMemoryAssignmentCache extends AbstractAssignmentCache {
+  public ExpiringInMemoryAssignmentCache(int cacheTimeout, TimeUnit timeUnit) {
+    super(Collections.synchronizedMap(new PassiveExpiringMap<>(cacheTimeout, timeUnit)));
+  }
+
+  public ExpiringInMemoryAssignmentCache(
```
typotter, line 14: @aarsilv Should these cache implementations be part of the `api` package since they'll be used downstream? or just part of the cache package since they'll be used in downstream only in the sdk and not in the developer's code.

aarsilv: Thinking cache package since they won't be directly used

src/main/java/cloud/eppo/BaseEppoClient.java
```diff
@@ -156,33 +165,55 @@ protected EppoValue getTypedAssignment(
...
+                extraLogging,
+                metaData);
+
+        // Deduplication of assignment logging is possible by providing an `IAssignmentCache`.
+        // Default to true, only avoid logging if there's a cache hit.
+        boolean logAssignment = true;
+        AssignmentCacheEntry cacheEntry = AssignmentCacheEntry.fromVariationAssignment(assignment);
+        if (assignmentCache != null) {
+          try {
+            if (assignmentCache.hasEntry(cacheEntry).get()) {
```
aarsilv, line 197: Given we call `get()` everywhere I'm thinking the cache interface should just be booleans and the use of completable futures can be done under the hood in concrete implementations

src/main/java/cloud/eppo/cache/ExpiringInMemoryAssignmentCache.java
```diff
@@ -0,0 +1,18 @@
+package cloud.eppo.cache;
+
+import cloud.eppo.api.AbstractAssignmentCache;
+import java.util.Collections;
+import java.util.Map;
+import java.util.concurrent.TimeUnit;
+import org.apache.commons.collections4.map.PassiveExpiringMap;
```
aarsilv, line 7: 🔥

src/test/java/cloud/eppo/BaseEppoClientBanditTest.java
```diff
@@ -66,14 +75,16 @@ public static void initClient() {
             false,
             false,
             true,
-            null);
+            null,
+            new AbstractAssignmentCache(assignmentCache) {},
+            new ExpiringInMemoryAssignmentCache(banditAssignmentCache, 10, TimeUnit.MINUTES) {});
```
aarsilv, line 80: 🔥 

I think for the test we should make it something like 50ms, then have the test sleep 50 MS, and make sure it logs the delayed duplicate assignment

aarsilv (approved):
Nice! I think we're in a good spot now. Approving, but with two comments I'd love for you to consider:
1. Our cache interface has methods that return completable futures. I think these could just be booleans, with futures pushed downstream
2. Our bandit logging test should exercise the expiring property of the expiring cache we pass it 

Take or leave either one as you see fit and then feel free to merge and release!

src/main/java/cloud/eppo/BaseEppoClient.java
typotter: You're right - this is actually a nice benefit of Java in that we can have functions that do the js equivalent of `await` without needing to also mark it as async.

PR #51, opened 2024-10-12
fix: make empty config default instead of null

typotter:
Eppo internal
🎟️ towards ff-3387

### DIFF
--- a/src/main/java/cloud/eppo/ConfigurationStore.java
+++ b/src/main/java/cloud/eppo/ConfigurationStore.java
@@ -7,18 +7,17 @@ import org.jetbrains.annotations.NotNull;
 /** Memory-only configuration store. */
 public class ConfigurationStore implements IConfigurationStore {
 
-  private volatile Configuration configuration;
+  // this is the fallback value if no configuration is provided (i.e. by fetch or initial config).
+  @NotNull private volatile Configuration configuration = Configuration.emptyConfig();
 
-  public ConfigurationStore() {
-    configuration = null;
-  }
+  public ConfigurationStore() {}
 
   public CompletableFuture<Void> saveConfiguration(@NotNull final Configuration configuration) {
     this.configuration = configuration;
     return CompletableFuture.completedFuture(null);
   }
 
-  public Configuration getConfiguration() {
+  @NotNull public Configuration getConfiguration() {
     return configuration;
   }
 }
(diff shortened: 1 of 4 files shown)


src/main/java/cloud/eppo/ConfigurationStore.java
```diff
@@ -7,18 +7,17 @@
...
-  }
+  public ConfigurationStore() {}
 
   public CompletableFuture<Void> saveConfiguration(@NotNull final Configuration configuration) {
     this.configuration = configuration;
     return CompletableFuture.completedFuture(null);
   }
 
-  public Configuration getConfiguration() {
+  @NotNull public Configuration getConfiguration() {
```
typotter, line 20: prevents other classes from ever having a null configuration.

src/main/java/cloud/eppo/IConfigurationStore.java
```diff
@@ -2,13 +2,14 @@
...
 import java.util.concurrent.CompletableFuture;
+import org.jetbrains.annotations.NotNull;
 
 /**
  * Common interface for extensions of this SDK to support caching and other strategies for
  * persisting configuration data across sessions.
  */
 public interface IConfigurationStore {
-  Configuration getConfiguration();
+  @NotNull Configuration getConfiguration();
```
typotter, line 12: strict contract that at least an empty configuration be returned by any implementations. Instead of having to special case "what if all the async loading fails", we simply enforce starting with an empty config, at least.

src/test/java/cloud/eppo/ConfigurationRequestorTest.java
```diff
@@ -21,7 +22,7 @@ public class ConfigurationRequestorTest {
 
   @Test
   public void testInitialConfigurationFuture() throws IOException {
-    IConfigurationStore configStore = new ConfigurationStore();
+    IConfigurationStore configStore = Mockito.spy(new ConfigurationStore());
```
typotter, line 25: in place of checking for a `null` configuration, spy on the config store and verify calls to `saveConfiguration` in addition to checking for empty

src/main/java/cloud/eppo/ConfigurationStore.java
```diff
@@ -7,18 +7,17 @@
 /** Memory-only configuration store. */
 public class ConfigurationStore implements IConfigurationStore {
 
-  private volatile Configuration configuration;
+  // this is the fallback value if no configuration is provided (i.e. by fetch or initial config).
+  @NotNull private volatile Configuration configuration = Configuration.emptyConfig();
```
aarsilv, line 11: 👍

aarsilv (approved):

PR #55, opened 2024-10-21
feat: extract minimal assignment cache map

typotter:
## Motivation and Context
The `AbstractAssignmentCache` contains an implementation of the contract specificed in `IAssignmentCache` that the cache entry's value be used when determining the presence of a given cache entry. To allow specific cache implementations of `IAssignmentCache` to make use of this existing implementation, we need to either reimplement Map<String, String> to pass as a delegate, or extract a minimal interface for cache delegates.

## Overview of Changes
-`CacheDelegate` interface for non map-based implementations
- retain the `protected AbstractAssignmentCache(final Map<String, String> delegate)` constructor and wrap the map in a `CacheDelegate` implementation.

### DIFF
--- a/src/main/java/cloud/eppo/api/AbstractAssignmentCache.java
+++ b/src/main/java/cloud/eppo/api/AbstractAssignmentCache.java
@@ -3,16 +3,42 @@ package cloud.eppo.api;
 import cloud.eppo.cache.AssignmentCacheEntry;
 import cloud.eppo.cache.AssignmentCacheKey;
 import java.util.Map;
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
 
 /**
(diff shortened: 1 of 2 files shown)


src/main/java/cloud/eppo/api/AbstractAssignmentCache.java
```diff
@@ -3,18 +3,44 @@
...
 
 /**
  * {@link IAssignmentCache} implementation which takes a map to use as the underlying storage
  * mechanism.
  */
 public abstract class AbstractAssignmentCache implements IAssignmentCache {
-  protected final Map<String, String> delegate;
 
-  protected AbstractAssignmentCache(final Map<String, String> delegate) {
+  /** Minimal "map" implementation required to store cached assignment data. */
```
typotter, line 15: @aarsilv I understand now :)

aarsilv (approved):
🎉

PR #68, opened 2024-11-15
fix: respect graceful mode during client initialization

typotter:
🎟️ [linear](https://linear.app/eppo/issue/FF-3565/implement-graceful-mode-error-handling-in-initialization)
towards FF-3565

## Motivation and Context
The SDK will throw an unchecked exception upon initialization/loading configuration if it is unable to fetch over the network; this can result in applications crashing. 

## Changes
- log an error if an exception is thrown during configuration load, but only re-throw exceptions if `gracefulMode=false`

## How is this tested
- unit tests simulating error at the HTTP level
- Unit test continuing to attempt assignment after failed fetch.

### DIFF
--- a/src/test/java/cloud/eppo/BaseEppoClientTest.java
+++ b/src/test/java/cloud/eppo/BaseEppoClientTest.java
@@ -2,18 +2,17 @@ package cloud.eppo;
 
 import static cloud.eppo.helpers.AssignmentTestCase.parseTestCaseFile;
 import static cloud.eppo.helpers.AssignmentTestCase.runTestCase;
+import static cloud.eppo.helpers.TestUtils.mockHttpError;
 import static cloud.eppo.helpers.TestUtils.mockHttpResponse;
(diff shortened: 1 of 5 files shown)


felipecsl (commented):
This is actually throwing a *checked* exception. Can we do unchecked instead? Checked exceptions are really annoying to work with

src/test/java/cloud/eppo/BaseEppoClientTest.java
```diff
@@ -66,7 +66,7 @@ private void initClientWithData(
         new BaseEppoClient(
             DUMMY_FLAG_API_KEY,
             isConfigObfuscated ? "android" : "java",
-            "3.0.0",
+            "100.1.0",
```
typotter, line 69: This version string does not need to equal what version the package is actually at. setting the test string to `100.1.0` makes it more clear than `3.0.0` which is close to actual.

felipecsl (approved):
nice! way better

PR #75, opened 2024-12-16
feat: use baseUrl instead of host in a backwards-compatible way

typotter:
towards FF-3558
🎟️ [FF-3558](https://linear.app/eppo/issue/FF-3558/unify-sdk-request-baseurl)

## Motivation and Context
Most, nearly all in fact, of the SDKs use a baseUrl of https://fscdn.eppo.cloud/api and endpoint constants of `/flag-config/v1/config` and `/flag-config/v1/bandits` for flag and bandit requests.
iOS and, at some point, the Java SDK [stopped following](https://github.com/Eppo-exp/sdk-common-jdk/blob/main/src/main/java/cloud/eppo/ConfigurationRequestor.java#L12) this [convention](https://github.com/Eppo-exp/sdk-common-jdk/blob/main/src/main/java/cloud/eppo/Constants.java) instead putting the /apiportion in the endpoint string.

All of the SDKs need to follow the common convention in order for our cross-SDK testing tools (scenario testing, benchmarking) to work seamlessly.

## Changes
- Add `baseUrl` parameter to constructor
- Allow either host or base parameter to be null, assigning the default if both are null
- prefer baseUrl but if null and host is supplied, append the expected 'api'


src/main/java/cloud/eppo/BaseEppoClient.java
```diff
@@ -56,7 +56,8 @@ protected BaseEppoClient(
       @NotNull String apiKey,
       @NotNull String sdkName,
       @NotNull String sdkVersion,
-      @Nullable String host,
+      @Nullable String apiBaseUrl,
+      @Deprecated @Nullable String apiHost, // To be removed in v4. use `apiBaseUrl` instead.
```
felipecsl, line 60: took me a minute to figure out that you renamed `host` to `apiHost` and deprecated it. I'd recommend undoing the rename (keep `host` to avoid confusion) and add the comment as standard javadocs above the method

felipecsl (approved):

src/main/java/cloud/eppo/BaseEppoClient.java
typotter: 👍

PR #89, opened 2025-02-21
feat: `getConfig` and `getFlagType`

typotter:
_Eppo Internal_
📜 [DD](https://www.notion.so/eppo/New-Java-SDK-methods-getConfiguration-and-getFlagType-19a49cc0114380709ff3e69f020650d8)
🎟️ Fixes FF-3454

## Motivation and Context
GlossGenius (and, likely other clients) integrate with the Eppo SDK in such a way that the type a given flag is not known at the time the assignment is requested from the SDK. In v2 of the SDK (Java server), `getAssignmentVariation` was available, [but removed in v3 to favour typed assignment returns.](https://eppo-group.slack.com/archives/C05FKKPDTJR/p1730229809171679?thread_ts=1730226618.797779&cid=C05FKKPDTJR) This method returned both the resulting value and its type. This information was used to convert the resulting value to the appropriate type for further processing.

There is a precedence for similar methods in other SDKs:

- the Rust SDK has [getAssignment](https://github.com/Eppo-exp/eppo-multiplatform/blob/cb8c6b441aaaf8ad2595caa76dddf9ea4097d84b/rust-sdk/src/client.rs#L112)
- The Ruby SDK exposes the [whole configuration object](https://github.com/Eppo-exp/eppo-multiplatform/blob/cb8c6b441aaaf8ad2595caa76dddf9ea4097d84b/ruby-sdk/lib/eppo_client/client.rb#L34) without an untyped getAssignment method
    - While Ruby/Python SDKs expose the configuration object, it’s not currently possible to programmatically get flag types from it. Configuration only has methods to get flags/bandits keys and to get serialized configuration
- Python also exposes the config through [get_configuration](https://github.com/Eppo-exp/eppo-multiplatform/blob/cb8c6b441aaaf8ad2595caa76dddf9ea4097d84b/python-sdk/src/client.rs#L428)
- JS has [getFlagConfigurations](https://github.com/Eppo-exp/js-sdk-common/blob/f6c3d4f2a1bb9e78538539fcfcd25ce1f44961a3/src/client/eppo-client.ts#L1154) which returns the raw DTO

The untyped assignment methods allow the assignment to be computed then converted by the developer into the correct type. The SDKs which expose the configuration require first a call to `getConfiguration().getFlagType(flagKey)`, then a call to the correct typed assignment method.

## Changes Overview
- expose the already existing `Configuration` object (which is immutable) via `getConfiguration`
- add a method to `getConfiguration` to return the variation type.

### DIFF
--- a/src/main/java/cloud/eppo/BaseEppoClient.java
+++ b/src/main/java/cloud/eppo/BaseEppoClient.java
@@ -203,7 +203,7 @@ public class BaseEppoClient {
     throwIfEmptyOrNull(flagKey, "flagKey must not be empty");
     throwIfEmptyOrNull(subjectKey, "subjectKey must not be empty");
 
-    Configuration config = configurationStore.getConfiguration();
+    Configuration config = getConfiguration();
 
     FlagConfig flag = config.getFlag(flagKey);
     if (flag == null) {
@@ -496,7 +496,7 @@ public class BaseEppoClient {
       Actions actions,
       String defaultValue) {
     BanditResult result = new BanditResult(defaultValue, null);
-    final Configuration config = configurationStore.getConfiguration();
+    final Configuration config = getConfiguration();
     try {
       String assignedVariation =
           getStringAssignment(
@@ -578,5 +578,9 @@ public class BaseEppoClient {
 
   public void setIsGracefulFailureMode(boolean isGracefulFailureMode) {
     this.isGracefulMode = isGracefulFailureMode;
+  }
+
+  public Configuration getConfiguration() {
+    return configurationStore.getConfiguration();
   }
 }
--- a/src/main/java/cloud/eppo/BaseEppoClient.java
+++ b/src/main/java/cloud/eppo/BaseEppoClient.java
@@ -581,7 +581,7 @@ public class BaseEppoClient {
   }
 
   /**
-   * Gets the configuration object used by the EppoClient for assignment and bandit evaluation.
+   * Returns the configuration object used by the EppoClient for assignment and bandit evaluation.
    *
    * <p>The configuration object is for debugging (inspect the loaded config) and other advanced use
    * cases where flag metadata or a list of flag keys, for example, is required.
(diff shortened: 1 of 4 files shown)


src/main/java/cloud/eppo/BaseEppoClient.java
```diff
@@ -579,4 +579,20 @@ private <T> T throwIfNotGraceful(Exception e, T defaultValue) {
   public void setIsGracefulFailureMode(boolean isGracefulFailureMode) {
     this.isGracefulMode = isGracefulFailureMode;
   }
+
+  /**
+   * Gets the configuration object used by the EppoClient for assignment and bandit evaluation.
```
felipecsl, line 584:
```suggestion
   * Returns the configuration object used by the EppoClient for assignment and bandit evaluation.
```

src/main/java/cloud/eppo/api/Configuration.java
```diff
@@ -121,6 +122,19 @@ public FlagConfig getFlag(String flagKey) {
     return flags.get(flagKeyForLookup);
   }
 
+  /**
+   * Gets the Variation Type for the specified flag if it exists, otherwise returns null.
```
felipecsl, line 126:
```suggestion
   * Returns the VariationType for the specified flag if it exists, otherwise returns null.
```

src/test/java/cloud/eppo/BaseEppoClientTest.java
```diff
@@ -655,6 +656,103 @@ public void testPolling() {
     eppoClient.stopPolling();
   }
 
+  @Test
+  public void testGetConfiguration() {
+    // Initialize client with default settings
+    initClient();
+
+    // Get configuration
+    Configuration config = eppoClient.getConfiguration();
```
felipecsl, line 665: theses comments smell like LLM 😂

felipecsl (approved):

src/test/java/cloud/eppo/BaseEppoClientTest.java
typotter: lol. ok, ya that one is particularly silly. The awkward doc comments are all mine though.

PR #102, opened 2025-04-04
feat: use subdomain encoded in SDK token 

typotter:
🎟️ 
Fixes FF-4276
Towards FF-4184
📜 [Design Doc](https://www.notion.so/eppo/Encoding-service-locations-in-SDK-keys-1bd49cc0114380c6ae1ac30fd0d42b5c?pvs=4)

## Motivation and Context
In order to support off-boarding of customers, we have created an architecture where requests to the CDN, Edge Compute, and Event Ingestion are all routed through customer-specific subdomains (which can be blocked when customers off-board.

### DIFF
--- a/src/main/java/cloud/eppo/ApiEndpoints.java
+++ b/src/main/java/cloud/eppo/ApiEndpoints.java
@@ -11,7 +11,7 @@ import org.jetbrains.annotations.Nullable;
  */
 public class ApiEndpoints {
 
-  private final SdkKey sdkToken;
+  private final SDKKey sdkToken;
   private final String baseUrl;
 
   /**
@@ -20,7 +20,7 @@ public class ApiEndpoints {
    * @param sdkKey SDK Key instance for subdomain
    * @param baseUrl Custom base URL (optional)
    */
-  public ApiEndpoints(@NotNull SdkKey sdkKey, @Nullable String baseUrl) {
+  public ApiEndpoints(@NotNull SDKKey sdkKey, @Nullable String baseUrl) {
     this.sdkToken = sdkKey;
     this.baseUrl = baseUrl;
   }
--- a/src/main/java/cloud/eppo/ApiEndpoints.java
+++ b/src/main/java/cloud/eppo/ApiEndpoints.java
@@ -46,7 +46,7 @@ public class ApiEndpoints {
     } else if (sdkKey.isValid()) {
       String subdomain = sdkKey.getSubdomain();
       if (subdomain != null) {
-        String domainPart = DEFAULT_BASE_URL.replaceAll("^(https?://|//)", "");
+        String domainPart = DEFAULT_BASE_URL.replaceAll("^(https?:)?//", "");
         effectiveUrl = subdomain + "." + domainPart;
       } else {
         effectiveUrl = DEFAULT_BASE_URL;
--- a/src/main/java/cloud/eppo/ApiEndpoints.java
+++ b/src/main/java/cloud/eppo/ApiEndpoints.java
@@ -11,7 +11,7 @@ import org.jetbrains.annotations.Nullable;
  */
 public class ApiEndpoints {
 
-  private final SDKKey sdkToken;
+  private final SDKKey sdkKey;
   private final String baseUrl;
 
   /**
@@ -21,13 +21,13 @@ public class ApiEndpoints {
    * @param baseUrl Custom base URL (optional)
    */
   public ApiEndpoints(@NotNull SDKKey sdkKey, @Nullable String baseUrl) {
-    this.sdkToken = sdkKey;
+    this.sdkKey = sdkKey;
     this.baseUrl = baseUrl;
   }
 
   /**
    * Gets the normalized base URL based on the following priority: 1. If baseUrl is provided and not
-   * equal to DEFAULT_BASE_URL, use it 2. If the SDK token contains a subdomain, use it with
+   * equal to DEFAULT_BASE_URL, use it 2. If the SDK Key contains a subdomain, use it with
    * DEFAULT_BASE_URL 3. Otherwise, fall back to DEFAULT_BASE_URL
    *
    * <p>The returned URL will: - Always have a protocol (defaults to https:// if none provided) -
@@ -41,10 +41,10 @@ public class ApiEndpoints {
     if (baseUrl != null && !baseUrl.equals(DEFAULT_BASE_URL)) {
       // This is to prevent forcing the SDK to send requests to the CDN server without a subdomain
       // even when encoded in
-      // SDK token.
+      // SDK Key.
       effectiveUrl = baseUrl;
-    } else if (sdkToken != null && sdkToken.isValid()) {
-      String subdomain = sdkToken.getSubdomain();
+    } else if (sdkKey.isValid()) {
+      String subdomain = sdkKey.getSubdomain();
       if (subdomain != null) {
         String domainPart = DEFAULT_BASE_URL.replaceAll("^(https?://|//)", "");
         effectiveUrl = subdomain + "." + domainPart;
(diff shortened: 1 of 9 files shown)


src/main/java/cloud/eppo/ApiEndpoints.java
```diff
@@ -0,0 +1,68 @@
...
+
+    if (baseUrl != null && !baseUrl.equals(DEFAULT_BASE_URL)) {
+      // This is to prevent forcing the SDK to send requests to the CDN server without a subdomain
+      // even when encoded in
+      // SDK token.
+      effectiveUrl = baseUrl;
+    } else if (sdkToken != null && sdkToken.isValid()) {
+      String subdomain = sdkToken.getSubdomain();
+      if (subdomain != null) {
+        String domainPart = DEFAULT_BASE_URL.replaceAll("^(https?://|//)", "");
```
aarsilv, line 49: I think this regex could be simplified to `^(https?:)?//` although not sure if that is more readable 😅

src/main/java/cloud/eppo/ApiEndpoints.java
```diff
@@ -0,0 +1,68 @@
...
+      effectiveUrl = DEFAULT_BASE_URL;
+    }
+
+    // Remove any trailing slashes
+    effectiveUrl = effectiveUrl.replaceAll("/+$", "");
+
+    // Add protocol if missing
+    if (!effectiveUrl.matches("^(https?://|//).*")) {
+      effectiveUrl = "https://" + effectiveUrl;
+    }
```
aarsilv, line 64: There is a lot of regex magic, which I'm OK with. However, I wonder if there is a first-class class (e.g., `URI`) we should be using instead.

src/main/java/cloud/eppo/SdkKey.java
```diff
@@ -0,0 +1,80 @@
...
+            pairParts.length > 1
+                ? URLDecoder.decode(pairParts[1], StandardCharsets.UTF_8.name())
+                : null;
+
+        query_pairs.put(key, value);
+      }
+      return query_pairs;
+
+    } catch (Exception e) {
+      return Collections.emptyMap();
```
aarsilv, line 54: How about logging a warning if we get here to aid in debugging?

src/main/java/cloud/eppo/SdkKey.java
```diff
@@ -0,0 +1,80 @@
...
+        return Collections.emptyMap();
+      }
+
+      String payload = parts[1];
+      if (payload == null) {
+        return Collections.emptyMap();
+      }
+
+      String decodedString = Utils.base64Decode(payload);
+      final Map<String, String> query_pairs = new LinkedHashMap<>();
```
aarsilv, line 35: Any reason for a `LinkedHashMap` vs. a regular `HashMap`? I don't see us iterating over this thing anywhere.

src/main/java/cloud/eppo/SdkKey.java
```diff
@@ -0,0 +1,80 @@
...
+    return sdkKey;
+  }
+
+  /**
+   * Checks if the SDK Key had the subdomain encoded.
+   *
+   * @return true if the token is valid and contains required parameters
+   */
+  public boolean isValid() {
+    return !decodedParams.isEmpty() && getSubdomain() != null;
```
aarsilv, line 78: I don't think we need this first check as if the params is empty then `getSubdomain()` will return null.

src/test/java/cloud/eppo/ApiEndpointsTest.java
```diff
@@ -0,0 +1,77 @@
+package cloud.eppo;
+
+import static org.junit.jupiter.api.Assertions.*;
+
+import org.junit.jupiter.api.Test;
+
+public class ApiEndpointsTest {
```
aarsilv, line 7: 🙌

src/test/java/cloud/eppo/ApiEndpointsTest.java
```diff
@@ -0,0 +1,77 @@
...
+  }
+
+  @Test
+  public void testCustomBaseUrl() {
+    ApiEndpoints endpoints = new ApiEndpoints(plainKey, "custom.domain/api");
+    assertEquals("https://custom.domain/api", endpoints.getBaseUrl());
+  }
+
+  @Test
+  public void testCustomBaseUrlWithProtocol() {
```
aarsilv, line 23: Should we have a test that includes a port?

src/test/java/cloud/eppo/SdkKeyTest.java
```diff
@@ -0,0 +1,90 @@
...
+
+    SdkKey decoder = new SdkKey(token);
+
+    assertTrue(decoder.isValid());
+    assertEquals("test-subdomain", decoder.getSubdomain());
+    assertEquals(token, decoder.getToken());
+  }
+
+  @Test
+  public void testTokenWithEncodedCharacters() {
```
aarsilv, line 68: 💪

src/test/java/cloud/eppo/SdkKeyTest.java
```diff
@@ -0,0 +1,90 @@
...
+
+    SdkKey decoder = new SdkKey(token);
+
+    assertTrue(decoder.isValid());
+    assertEquals("test subdomain", decoder.getSubdomain());
+    assertEquals(token, decoder.getToken());
+  }
+
+  @Test
+  public void testTokenWithMalformedBase64() {
```
aarsilv, line 81: 💪

aarsilv (approved):
Great stuff! Clean, and well tested 💪 

Left a bunch of minor comments you can take or leave as you see fit!

src/main/java/cloud/eppo/ApiEndpoints.java
```diff
@@ -0,0 +1,68 @@
...
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
+/**
+ * Utility class for constructing Eppo API base URL. Determines the effective base URL considering
+ * baseUrl, subdomain from SDK token, and defaultUrl in that order.
+ */
+public class ApiEndpoints {
+
+  private final SdkKey sdkToken;
```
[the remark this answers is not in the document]
typotter: good catch. the terms are not properly precise here.
The SDKKey object is a thing that is identified by the SDK token. Of course, the SDK key is just a special kind of API Key. There's good reason for all the interchanging in the code, but we can be more precise now (especially that I understand the genealogy now).

src/main/java/cloud/eppo/ApiEndpoints.java
typotter: There is! Except the `javax.net` class is not available in Android and `android.net` is not available outside. 😢

src/main/java/cloud/eppo/SdkKey.java
typotter: done

src/main/java/cloud/eppo/SdkKey.java
typotter: not particulary

src/main/java/cloud/eppo/SdkKey.java
typotter: I think that actually, I will drop the subdomain check from `isValid`. We always check the subdomain presence before using it anyway.

src/test/java/cloud/eppo/ApiEndpointsTest.java
typotter: yes def

src/main/java/cloud/eppo/ApiEndpoints.java
typotter: done

PR #109, opened 2025-04-15
[Draft] Android api 21 support

hbordersTwitch:
This is a draft of implementations necessary to run on Android API 21. I haven't begun actually testing this on an Android API 21 emulator yet because I still need to work through backporting [android-sdk](https://github.com/Eppo-exp/android-sdk/), but this should give you an idea of the direction I'm going.

Major issues:
1. `String.join()` requires JDK8, so use StringBuilder instead
2. `java.util.Base64` requires JDK8, so use `net.iharder:base64` (Public Domain) instead
3. Operations using fork/join pools and `CompletableFuture` require JDK8, so create blocking versions of these classes, and have Android API 21 consumers handle concurrency manually.
4. `Jackson` requires JDK7, so put all Jackson consumption behind an interface so that Android API 21 can inject a `JSONObject`-based implementation.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -20,6 +20,8 @@ dependencies {
   implementation 'com.fasterxml.jackson.core:jackson-databind:2.18.3'
   implementation 'com.github.zafarkhaja:java-semver:0.10.2'
   implementation "com.squareup.okhttp3:okhttp:4.12.0"
+  // For base64 that works on on Android API 21
+  implementation "net.iharder:base64:2.3.8"
   // For LRU and expiring maps
   implementation 'org.apache.commons:commons-collections4:4.4'
   implementation 'org.slf4j:slf4j-api:2.0.17'
--- a/src/main/java/cloud/eppo/BaseEppoClient.java
+++ b/src/main/java/cloud/eppo/BaseEppoClient.java
@@ -6,6 +6,12 @@ import static cloud.eppo.Utils.throwIfEmptyOrNull;
 
 import cloud.eppo.api.*;
 import cloud.eppo.cache.AssignmentCacheEntry;
+import cloud.eppo.json.JacksonMapper;
+import cloud.eppo.json.JacksonMapperNode;
+import cloud.eppo.json.Mapper;
+import cloud.eppo.json.MapperException;
+import cloud.eppo.json.MapperJsonProcessingException;
+import cloud.eppo.json.MapperNode;
 import cloud.eppo.logging.Assignment;
 import cloud.eppo.logging.AssignmentLogger;
 import cloud.eppo.logging.BanditAssignment;
@@ -27,9 +33,7 @@ import org.slf4j.LoggerFactory;
 
 public class BaseEppoClient {
   private static final Logger log = LoggerFactory.getLogger(BaseEppoClient.class);
-  private final ObjectMapper mapper =
-      new ObjectMapper()
-          .registerModule(EppoModule.eppoModule()); // TODO: is this the best place for this?
+  private final Mapper mapper;
 
   protected final ConfigurationRequestor requestor;
 
@@ -73,6 +77,46 @@ public class BaseEppoClient {
       @Nullable CompletableFuture<Configuration> initialConfiguration,
       @Nullable IAssignmentCache assignmentCache,
       @Nullable IAssignmentCache banditAssignmentCache) {
+    this(
+        new JacksonMapper(),
+        apiKey,
+        sdkName,
+        sdkVersion,
+        host,
+        apiBaseUrl,
+        assignmentLogger,
+        banditLogger,
+        configurationStore,
+        isGracefulMode,
+        expectObfuscatedConfig,
+        supportBandits,
+        initialConfiguration,
+        assignmentCache,
+        banditAssignmentCache
+    );
+  }
+
+  // It is important that the bandit assignment cache expire with a short-enough TTL to last about
+  // one user session.
+  // The recommended is 10 minutes (per @Sven)
+  /** @param host To be removed in v4. use `apiBaseUrl` instead. */
+  protected BaseEppoClient(
+      @NotNull Mapper mapper,
+      @NotNull String apiKey,
+      @NotNull String sdkName,
+      @NotNull String sdkVersion,
+      @Deprecated @Nullable String host,
+      @Nullable String apiBaseUrl,
+      @Nullable AssignmentLogger assignmentLogger,
+      @Nullable BanditLogger banditLogger,
+      @Nullable IConfigurationStore configurationStore,
+      boolean isGracefulMode,
+      boolean expectObfuscatedConfig,
+      boolean supportBandits,
+      @Nullable CompletableFuture<Configuration> initialConfiguration,
+      @Nullable IAssignmentCache assignmentCache,
+      @Nullable IAssignmentCache banditAssignmentCache) {
+    this.mapper = mapper;
 
     if (apiKey == null) {
       throw new IllegalArgumentException("Unable to initialize Eppo SDK due to missing API key");
@@ -436,7 +480,12 @@ public class BaseEppoClient {
               subjectAttributes,
               EppoValue.valueOf(defaultValue.toString()),
               VariationType.JSON);
-      return parseJsonString(value.stringValue());
+      MapperNode mapperNode = parseJsonString(value.stringValue());
+      if (mapperNode != null) {
+        return ((JacksonMapperNode) mapperNode).getJsonNode();
+      } else {
+        return null;
+      }
     } catch (Exception e) {
       return throwIfNotGraceful(e, defaultValue);
     }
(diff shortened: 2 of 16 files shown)


src/test/java/cloud/eppo/ConfigurationRequestorJava6Test.java
```diff
@@ -0,0 +1,424 @@
...
+        } while (serverSocket == null && port < endPort);
+        if (serverSocket == null) {
+            fail("Couldn't allocate ServerSocket between [" + startPort + ", " + endPort + ")");
+        }
+
+        return serverSocket;
+    }
+
+    @Test
+    public void testInterruptedFetchDoesNotClobberCache() throws Exception {
```
hbordersTwitch, line 215: An equivalent to this test didn't previously exist, but I think it's important to verify that thread interruption works since there needs to be a way to cancel blocking operations.

build.gradle
```diff
@@ -20,6 +20,8 @@ dependencies {
   implementation 'com.fasterxml.jackson.core:jackson-databind:2.18.3'
   implementation 'com.github.zafarkhaja:java-semver:0.10.2'
   implementation "com.squareup.okhttp3:okhttp:4.12.0"
+  // For base64 that works on on Android API 21
+  implementation "net.iharder:base64:2.3.8"
```
aarsilv, line 24: I think dropping in a mutually compatible library is fine. However, we faced a similar challenge with PHP and caching, and opted to have the upstream SDK essentially pass in the library. I wonder if we should consider doing similar here, basically having this SDK take an interface that can Base64 encode/decode and then Java SDK will pass in something that uses `java.util.Base64` and Android SDK something that uses `android.util.Base64`?

Note it seems you're doing something similar for JSON mapping

src/main/java/cloud/eppo/BaseEppoClient.java
```diff
@@ -436,7 +480,12 @@ public JsonNode getJSONAssignment(
               subjectAttributes,
               EppoValue.valueOf(defaultValue.toString()),
               VariationType.JSON);
-      return parseJsonString(value.stringValue());
+      MapperNode mapperNode = parseJsonString(value.stringValue());
+      if (mapperNode != null) {
+        return ((JacksonMapperNode) mapperNode).getJsonNode();
```
aarsilv, line 485: Would we want to stay away from Jackson-specific stuff here?

I vaguely recall one of the main reasons we used Jackson is because it could more easily differentiate between arrays and objects than Java's built-in JSON parsing. But if we code around that, we may be able to remove Jackson entirely. Unless it's much slower on parsing, which may matter for Android app initialization times. In which case, I think upstream-provided parser as you're doing here may be the move.

src/main/java/cloud/eppo/ConfigurationRequestorJava6.java
```diff
@@ -0,0 +1,99 @@
...
+import org.slf4j.LoggerFactory;
+
+import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.concurrent.atomic.AtomicReference;
+import java.util.function.Consumer;
+
+import cloud.eppo.api.Configuration;
+import cloud.eppo.callback.CallbackManager;
+
+public class ConfigurationRequestorJava6 {
```
aarsilv, line 14: To help maximize code reuse and prevent drift, what if this extends the non-java-6 flavor and then uses the template pattern, with only the language level differences overwritten in the `ConfigurationRequestorJava6` class?

Note: I don't know if this is possible for building. A quick chat with AI suggests its not, and what you have is the way to go 😞

src/main/java/cloud/eppo/api/EppoValue.java
```diff
@@ -101,7 +101,17 @@ public String toString() {
...
-        return String.join(" ,", this.stringArrayValue);
+        if (this.stringArrayValue.isEmpty()) {
+          return "";
+        } else {
+          StringBuilder stringBuilder = new StringBuilder();
+          String delimiter = " ,";
+          for (String string : this.stringArrayValue) {
+            stringBuilder.append(string).append(delimiter);
+          }
+          stringBuilder.setLength(stringBuilder.length() - delimiter.length());
```
aarsilv, line 112: Ah this is to avoid a trailing delimiter 👌

src/test/java/cloud/eppo/ConfigurationRequestorJava6Test.java
```diff
@@ -0,0 +1,424 @@
...
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.concurrent.atomic.AtomicInteger;
+import java.util.concurrent.atomic.AtomicReference;
+
+import cloud.eppo.api.Configuration;
+
+public class ConfigurationRequestorJava6Test {
```
aarsilv, line 38: I wonder if to remove code duplication and keep tests in sync we could have one test suite that using a data provider or something runs each test on both implementations. Although things may get funky with mocking, but perhaps we could have an abstract parent class or shared helper methods or something.

aarsilv (commented):
Thanks for engaging on this! Things are making sense to me at a high-level, making the ability to drop in compatible internals where needed as well creating compatible flavors of those internals.

For encoding, I wonder if we just fully delegate that to upstream.

For json parsing its trickier, as Jackson will likely be faster something we want for both Android and Java. However, will Api level 21 compile with it present in the codebase? And if so, could we do something like the template pattern to reduce duplication? The latter is my main concern, the duplication could lead to drift between Java6 and "normal" flavors of things (and their tests!).

I'm less familiar with multi-API-level programming practices as well as the available annotations and their effects, so apologies if some of what I say does not make sense.

src/test/java/cloud/eppo/ConfigurationRequestorJava6Test.java
hbordersTwitch: I could create some abstractions to reduce duplication, but I find that too much abstraction in tests obscures the point of the test itself, so I'm generally ok with copypasta in test-land.

I also like this as a separate class because it establishes strong boundaries between Java6-land and Java8-land.

However, if this doesn't match the philosophy of your project, I'm happy to refactor.

src/main/java/cloud/eppo/api/EppoValue.java
hbordersTwitch: I could also do:
```
for (Iterator<String> i = this.stringArrayValue; i.hasNext();) {
  String string = i.next();
  stringBuilder.append(string);
  if (i.hasNext()) {
    stringBuilder.append(delimiter);
  }
}
```

src/main/java/cloud/eppo/ConfigurationRequestorJava6.java
typotter: Would we end up in callback hell if we just refactored the core to use callbacks instead of `CompletableFuture`s? The initialization code is certainly cleaner with `CompletableFuture`s

typotter (commented):
Thank you for tackling this! It's a tough set of use cases to make work all while respecting the API

As mentioned, I'd really like to get away from code duplication as much as possible. 
If we just convert all the `CompletableFuture`s to callbacks, for instance, how much does that simplify the implementation? Do you feel limited by respecting the API and not introducing a breaking change?

build.gradle
hbordersTwitch: That sounds good to me. We're still investigating whether we want to modify the Eppo SDK further, so I won't make this update yet, but if we decide to press forward, I'll definitely incorporate this change.

src/main/java/cloud/eppo/ConfigurationRequestorJava6.java
hbordersTwitch: I'd actually like to get rid of `CompletableFuture` if possible and provide callbacks because `CompletableFuture` relies on the ForkJoin Pool, and clients would probably like to minimize the number of thread pools in their apps.

You could certainly provide wrapper APIs that expose `CompletableFuture`, but some clients are going to use `RxJava` (which has its own thread pools), and some will use Kotlin coroutines (which has its own thread pools). I imagine customers would like to avoid creating one more thread at startup if possible.

A simple way to do this would be to provide a synchronous API like I did here, and then provide a `CompletableFuture` wrapper around it.

hbordersTwitch:
> Do you feel limited by respecting the API and not introducing a breaking change?

Yes. I was really trying not to introduce any breaking changes into your API. I was also trying to keep the Java6 implementation very separate to make it easy to rip back out again One Day.

typotter:
> Yes. I was really trying not to introduce any breaking changes into your API. I was also trying to keep the Java6 implementation very separate to make it easy to rip back out again One Day.
I appreciate you thinking about our codebase longterm and giving it the care you'd give your own repo.

If we have to introduce some breaking changes to eliminate developer friction, I'm all for it. Especially in the common SDK where we can adapt as needed in the server and android SDKs.

src/main/java/cloud/eppo/ConfigurationRequestorJava6.java
hbordersTwitch: I think callbacks make the most sense, and then if you wanted to go for extra credit, you could publish extension libraries with RxJava2, RxJava3, Kotlin Coroutines, and CompletableFutures bindings.

Modern Android developers are mostly using Kotlin Coroutines now. RxJava2 was common until about 2020, so there are probably still a lot of apps out there using it. I don't think RxJava3 is very common because it came out around the same time as Kotlin Coroutines.

hbordersTwitch:
We updated our minSDK to 26, so we no longer need this.

PR #111, opened 2025-04-22
chore: replace use of string.join for android 21

typotter:
_Eppo Internal:_

:tickets: Fixes FF-4349

## Motivation and Context
There are a number of incompatible libraries/language features/methods in java that don't work universally across java versions, server vs android deployments and across Android APIs. `String.join` is one such method.

## Description
- bespoke implementation of string join
 
n/a

## How has this been tested?
existing tests and new unit tests.


src/main/java/cloud/eppo/api/EppoValue.java
```diff
@@ -130,4 +132,20 @@ public boolean equals(Object otherObject) {
   public int hashCode() {
     return Objects.hash(type, boolValue, doubleValue, stringValue, stringArrayValue);
   }
+
+  private static String joinStringArray(List<String> stringArray) {
+    if (stringArray == null || stringArray.isEmpty()) {
+      return "";
+    }
+    String delimiter = " ,";
```
typotter, line 140: Any idea why we join with `" ,"` instead of `", "`?

aarsilv: Good spot! I think it was probably a mistake made earlier; I think we use string arrays for ONE OF checks and don't rely on `toString()` so you can fix this up.

src/main/java/cloud/eppo/api/EppoValue.java
```diff
@@ -130,4 +132,21 @@ public boolean equals(Object otherObject) {
...
     return Objects.hash(type, boolValue, doubleValue, stringValue, stringArrayValue);
   }
+
+  /** This method is to allow for Android 21 support; String.join was introduced in API 26 */
+  private static String joinStringArray(List<String> stringArray) {
+    if (stringArray == null || stringArray.isEmpty()) {
+      return "";
+    }
+    String delimiter = " ,";
+    StringBuilder stringBuilder = new StringBuilder();
```
aarsilv, line 142: 👍

src/test/java/cloud/eppo/api/EppoValueTest.java
```diff
@@ -0,0 +1,63 @@
+package cloud.eppo.api;
+
+import static org.junit.jupiter.api.Assertions.*;
+
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+import org.junit.jupiter.api.Test;
+
+public class EppoValueTest {
```
aarsilv, line 10: We actually have an `EppoValueTest` already, but it was left in `src/test/java/cloud/eppo/` (vs. API).
How about moving it here and then merging in the below test cases?

src/test/java/cloud/eppo/api/EppoValueTest.java
```diff
@@ -0,0 +1,63 @@
...
+
+  @Test
+  public void testToStringWithStringArray() {
```
aarsilv, line 13: 💪

aarsilv (approved):
Thanks for helping potentially support a lower API level! Agree on fixing the comma placement.

Approving assuming you merge the two `EppoValueTest` classes into one 💪

src/main/java/cloud/eppo/api/EppoValue.java
typotter: ✔️

src/test/java/cloud/eppo/api/EppoValueTest.java
typotter: ✔️

PR #112, opened 2025-04-24
chore: v4 - android21 compatible mega PR

typotter:
_Eppo Internal:_

:tickets: Fixes FF-4351

## Motivation and Context
In order to support Android API level 21, we need to remove some of the more modern features previously integrated. One such construct is the `CompletableFuture`, java's version of the `Promise`. In order to be compatible all the way back to java6, we need to replace these async structures with Callbacks for better or worse (and we can't even use lambdas)

## Description
- Change the `CallbackManager` class to use a dispatcher that gets around using different callback structures with a callback manager (needed only because we can't use lambdas)
- remove all `CompletableFuture`s, plumb callbacks everywhere for async
- Extracted interface for HttpClient
- new MockHttpClient classes for greatly simplifying tests.

**Breaking Changes**
- Initial Config can no longer be a future.
- Signature changed: `protected void loadConfigurationAsync(EppoActionCallback<Configuration> callback)`
- Signature changed: `public void saveConfiguration(@NotNull final Configuration configuration)`
- renamed `apiKey` to `sdkKey`
- removed deprecated `expectObfuscated` parameter
- removed deprecated `host` parameter
- removed deprecated Configuration.Builder method

**New API**
- New method: `activateConfiguration`

[//]: # (Please describe how you documented the developer impact of your changes; link to PRs or issues or explain why no documentation changes are required)
- [ ] Update docs as needed in downstream SDKs

## How has this been tested?
- [ ] Integrate in downstream SDKs and test

### DIFF
--- a/src/main/java/cloud/eppo/api/Configuration.java
+++ b/src/main/java/cloud/eppo/api/Configuration.java
@@ -110,14 +110,20 @@ public class Configuration {
 
   @Override
   public String toString() {
-    return "Configuration{" +
-      "banditReferences=" + banditReferences +
-      ", flags=" + flags +
-      ", bandits=" + bandits +
-      ", isConfigObfuscated=" + isConfigObfuscated +
-      ", flagConfigJson=" + Arrays.toString(flagConfigJson) +
-      ", banditParamsJson=" + Arrays.toString(banditParamsJson) +
-      '}';
+    return "Configuration{"
+        + "banditReferences="
+        + banditReferences
+        + ", flags="
+        + flags
+        + ", bandits="
+        + bandits
+        + ", isConfigObfuscated="
+        + isConfigObfuscated
+        + ", flagConfigJson="
+        + Arrays.toString(flagConfigJson)
+        + ", banditParamsJson="
+        + Arrays.toString(banditParamsJson)
+        + '}';
   }
 
   @Override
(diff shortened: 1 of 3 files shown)


src/main/java/cloud/eppo/callback/CallbackManager.java
```diff
@@ -3,7 +3,7 @@
 import java.util.Map;
 import java.util.UUID;
 import java.util.concurrent.ConcurrentHashMap;
-import java.util.function.Consumer;
+import org.jetbrains.annotations.NotNull;
```
aarsilv, line 6: At first I was thinking why have JetBrains bleed into the source, but then I remembered Android Studio is based off JetBrains 😅

aarsilv (approved):
Way to puzzle through this! 💪 💪 💪 

I imagine for the upstream clients we could still present the completable future interface if desired by wrapping the callbacks. Also Lambdas may be ok with "desugaring".

src/main/java/cloud/eppo/ConfigurationRequestor.java
```diff
@@ -1,84 +1,47 @@
...
 
-  private CompletableFuture<Void> remoteFetchFuture = null;
-  private CompletableFuture<Boolean> configurationFuture = null;
-  private boolean initialConfigSet = false;
-
-  private final CallbackManager<Configuration> configChangeManager = new CallbackManager<>();
+  private final CallbackManager<Configuration, Configuration.ConfigurationCallback>
+      configChangeManager =
+          new CallbackManager<>(
+              // no lambdas before java8
```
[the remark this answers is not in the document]
typotter: The handy thing is that for the dev, if they're using a platform that supports lambdas, they can use a lambda when they register a callback (this is in `EppoClient`). Since the callback interface there is a once function interface, Java will treat it interchangeably with a lambda whose signature matches the function.

src/main/java/cloud/eppo/callback/CallbackManager.java
typotter: It's just a little harmless documentation and contract declaration

PR #117, opened 2025-05-09
chore-feat: refactor out base64 ops

typotter:
_Eppo Internal:_

:tickets: Fixes FF-4351

## Motivation and Context
`java.util.Base64` is not available in Android21, but `android.util.base64` is. 

## Description
- Base64codec interface
- `Utils` static class gets a static codec reference and throws if not set
- Tests in the common SDK need to set the codec in a @BeforeAll method if the code under test requires base64 encoding/decoding
- Downstream SDKs must set an implementation of the interface which can be done statically in the EppoClient class as shown below.

Downstream (and test) implementations

```java
public class EppoClient extends BaseEppoClient {
  // Provide a base64 codec based on Androids base64 util.
  static {
    Utils.base64Codec =
        new Utils.Base64Codec() {
          @Override
          public String base64Encode(String input) {
            return new String(
                Base64.encode(input.getBytes(StandardCharsets.UTF_8), Base64.DEFAULT),
                StandardCharsets.UTF_8);
          }

          @Override
          public String base64Decode(String input) {
            return new String(Base64.decode(input, Base64.DEFAULT), StandardCharsets.UTF_8);
          }
        };
  }
```


src/main/java/cloud/eppo/Utils.java
```diff
@@ -16,6 +14,7 @@ public final class Utils {
   private static final SimpleDateFormat UTC_ISO_DATE_FORMAT = buildUtcIsoDateFormat();
   private static final Logger log = LoggerFactory.getLogger(Utils.class);
   private static final MessageDigest md = buildMd5MessageDigest();
+  public static Base64Codec base64Codec;
```
aarsilv, line 17: How about making this private with a setter

aarsilv (approved):

src/main/java/cloud/eppo/Utils.java
typotter: done

PR #124, opened 2025-05-26
chore: migrate to Maven Central for publishing

typotter:
We were previously using the Sonatype01 (s01) to publish to Maven Central. Sonatype is being turned down and namespaces migrated to Maven Central. 

Here, we update the server endpoints for our artifacts, continuing to use the `gradle publish` plugin, but with MAVEN credentials


.github/workflows/publish-sdk.yml
```diff
@@ -5,8 +5,8 @@ on:
     types: [published]
 
 env:
-  ORG_GRADLE_PROJECT_ossrhUsername: ${{ secrets.OSSRH_USERNAME }}
-  ORG_GRADLE_PROJECT_ossrhPassword: ${{ secrets.OSSRH_PASSWORD }}
+  ORG_GRADLE_PROJECT_ossrhUsername: ${{ secrets.MAVEN_CENTRAL_TOKEN_USERNAME }}
```
typotter, line 8: Requires these secrets to be set on the repo.

aarsilv: Did you have sufficient privileges to set them or do you need to pair?

build.gradle
```diff
@@ -109,8 +109,8 @@ publishing {
   }
   repositories {
     maven {
-      def releasesRepoUrl = "https://s01.oss.sonatype.org/service/local/staging/deploy/maven2/"
-      def snapshotsRepoUrl = "https://s01.oss.sonatype.org/content/repositories/snapshots/"
+      def releasesRepoUrl = "https://oss.sonatype.org/service/local/staging/deploy/maven2/"
+      def snapshotsRepoUrl = "https://oss.sonatype.org/content/repositories/snapshots/"
```
aarsilv, line 113: 📈

aarsilv (approved):

.github/workflows/publish-sdk.yml
typotter: secrets are set, thank you

PR #127, opened 2025-06-05
Added equals/hashCode/toString methods to all bean objects

hbordersTwitch:
_Eppo Internal:_

:tickets: Fixes __issue__
:scroll: Design Doc: __link if applicable__

## Motivation and Context
This makes unit testing easier for consumers. It shouldn't change any behaviors.

## Description
I generated everything with Android Studio. The only places I saw equals/hashCode/toString already had them directly after constructors, and ordered them: toString
equals
hashCode

so I preserved that position and order (and moved AssignmentCacheEntry's existing equals/hashCode to make it match that order).

No new documentation. This is just standard bean behavior

## How has this been tested?
Your existing unit tests should suffice

### DIFF
--- a/src/main/java/cloud/eppo/cache/AssignmentCacheEntry.java
+++ b/src/main/java/cloud/eppo/cache/AssignmentCacheEntry.java
@@ -27,6 +27,27 @@ public class AssignmentCacheEntry {
         new BanditCacheValue(assignment.getBandit(), assignment.getAction()));
   }
 
+  @Override
+  public boolean equals(Object o) {
+    if (o == null || getClass() != o.getClass()) return false;
+    AssignmentCacheEntry that = (AssignmentCacheEntry) o;
+    return Objects.equals(key, that.key)
+      && Objects.equals(value, that.value);
+  }
+
+  @Override
+  public int hashCode() {
+    return Objects.hash(key, value);
+  }
+
+  @Override
+  public String toString() {
+    return "AssignmentCacheEntry{" +
+      "key=" + key +
+      ", value=" + value +
+      '}';
+  }
+
   @NotNull public AssignmentCacheKey getKey() {
     return key;
   }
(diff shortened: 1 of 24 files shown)


src/main/java/cloud/eppo/cache/AssignmentCacheEntry.java
```diff
@@ -43,17 +64,4 @@ public static AssignmentCacheEntry fromBanditAssignment(BanditAssignment assignm
     return value;
   }
 
-  @Override
-  public boolean equals(Object o) {
-    if (this == o) return true;
```
typotter, line 48: Is this not still a useful check for equivalency?

typotter (approved):
Thanks @hbordersTwitch!

src/main/java/cloud/eppo/cache/AssignmentCacheEntry.java
hbordersTwitch: It is. I just moved the method to be consistent with the where you had the methods in other bean classes.

src/main/java/cloud/eppo/cache/AssignmentCacheEntry.java
```diff
@@ -27,6 +27,27 @@ public static AssignmentCacheEntry fromBanditAssignment(BanditAssignment assignm
         new BanditCacheValue(assignment.getBandit(), assignment.getAction()));
   }
 
+  @Override
+  public boolean equals(Object o) {
+    if (o == null || getClass() != o.getClass()) return false;
+    AssignmentCacheEntry that = (AssignmentCacheEntry) o;
+    return Objects.equals(key, that.key)
```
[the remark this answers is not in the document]
hbordersTwitch: We discussed this in a separate medium, and we want to revert this change.

PR #128, opened 2025-06-06
Thread-safe md5

aarsilv:
_Eppo Internal:_
:tickets: **Ticket:** FFL-163 - [[Legacy Eppo][Common Java SDK] Thread-safe md5](https://datadoghq.atlassian.net/browse/FFL-163?atlOrigin=eyJpIjoiNzcwNDk3OWY5ZjE2NGFmOWI2N2QxYmZjZmYxMDdlZTkiLCJwIjoiaiJ9)
🗨️ **Slack:** ["...multiple threads add to the digest between resets ..."](https://eppo-group.slack.com/archives/C06EL6BM4UT/p1749152885395699?thread_ts=1749129344.545099&cid=C06EL6BM4UT)

## Motivation and Context
When refreshing configurations in multiple threads, errors like `Unknown operator "27457ce369f2a74203396a35ef537c0b"` were being encountered. This operator _is_ known, so what is likely happening is getting the MD5 hash--formerly used by done by Operator deserialization--is including multiple strings when digesting.

## Description
First I reproduced the issue with a unit test.
Then we made the static `MessageDigest` a `ThreadLocal` so it's not shared between threads, and the test passed.
Lastly, we took the opportunity to clean up Operator deserialization from strings so that it doesn't need to compute MD5 hashes at all, which should be a performance gain.

_(no documentation updates needed)_

## How has this been tested?
* Added a `testGetMd5HashThreadSafe()` method to `UtilsTest`
* Existing `BaseEppoClientTest`

### DIFF
--- a/src/test/java/cloud/eppo/UtilsTest.java
+++ b/src/test/java/cloud/eppo/UtilsTest.java
@@ -32,7 +32,7 @@ public class UtilsTest {
   public void testGetMd5HashThreadSafe() {
     final AtomicBoolean interferenceEncountered = new AtomicBoolean(false);
     int numThreads = 2;
-    ExecutorService pool = Executors.newFixedThreadPool(2);
+    ExecutorService pool = Executors.newFixedThreadPool(numThreads);
     try {
       for (int i = 0; i < numThreads; i += 1) {
         pool.execute(


src/test/java/cloud/eppo/UtilsTest.java
```diff
@@ -25,6 +28,42 @@ public void testGetMd5Hash() {
...
+        fail();
+      }
+      assertFalse(interferenceEncountered.get());
+    }
+  }
+
+  private boolean testForMd5Interference() {
+    boolean interferenceEncountered = false;
+    for (int i = 0; i < 100; i += 1) {
+      if (!getMD5Hex("input-62").equals("448614887a99f16179b400cfccceb72d")) {
```
sameerank, line 59: 👍 In other words, in every thread, the hash must be the same. Interference would mean appending to the hashed input -> different hash outputs

sameerank (approved):
In the tests I trust

src/main/java/cloud/eppo/ufc/dto/OperatorType.java
```diff
@@ -14,20 +17,40 @@ public enum OperatorType {
...
   public static OperatorType fromString(String value) {
-    for (OperatorType type : OperatorType.values()) {
-      if (type.value.equals(value)
-          || getMD5Hex(type.value).equals(value)
-          || getMD5Hex(type.value).equals(value)) {
-        return type;
-      }
+    // First we try obfuscated lookup as in client situations we'll care more about ingestion
+    // performance
+    OperatorType type = hashesToOperatorType.get(value);
```
aarsilv, line 48: We really shouldn't be hashing operators more than once so as a performance improvement we remove the loop and just built two maps for unhashed and hashed operators on init.

src/test/java/cloud/eppo/UtilsTest.java
aarsilv: Exactly! If two threads add to the digest at the same time between `reset()` calls they both get bad hashes.

PR #149, opened 2025-09-04
Added BaseEppoClient.getTypedAssignmentResult + nullability

hbordersTwitch:
_Eppo Internal:_

:tickets: Fixes __issue__
:scroll: Design Doc: __link if applicable__

## Motivation and Context
When evaluating a feature flag, I want to know if the evaluation failed and why. We report the state of all of our feature flags into Crashlytics as we evaluate them, and if a feature flag failed to load, we want to update its state accordingly.

## Description
Added minimum nullability necessary to add required fields. Also added `sortedVariationKeys` to `FlagConfig` to make finding the variation index cheaper.

If you strongly oppose the nullability annotations, I can remove them, but it made reasoning about the code much simpler, and I found a couple of missing null checks because of them.

Added documentation to new classes.

## How has this been tested?
Added unit tests. Not sure how to unit test no allocations inside `BaseEppoClientTest`

### DIFF
--- a/src/main/java/cloud/eppo/api/EppoValue.java
+++ b/src/main/java/cloud/eppo/api/EppoValue.java
@@ -1,123 +1,184 @@
 package cloud.eppo.api;
 
+import static cloud.eppo.Utils.throwIfNull;
+
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
 import cloud.eppo.ufc.dto.EppoValueType;
-import java.util.Iterator;
 import java.util.List;
 import java.util.Objects;
 
 public class EppoValue {
-  protected final EppoValueType type;
-  protected Boolean boolValue;
-  protected Double doubleValue;
-  protected String stringValue;
-  protected List<String> stringArrayValue;
+  @NotNull protected final EppoValueType type;
+  @Nullable protected final Boolean boolValue;
+  @Nullable protected final Double doubleValue;
+  @Nullable protected final String stringValue;
+  @Nullable protected final List<String> stringArrayValue;
 
   protected EppoValue() {
-    this.type = EppoValueType.NULL;
+    this(
+      EppoValueType.NULL,
+      null,
+      null,
+      null,
+      null
+    );
   }
 
   protected EppoValue(boolean boolValue) {
-    this.boolValue = boolValue;
-    this.type = EppoValueType.BOOLEAN;
+    this(
+      EppoValueType.BOOLEAN,
+      boolValue,
+      null,
+      null,
+      null
+    );
   }
 
   protected EppoValue(double doubleValue) {
+    this(
+      EppoValueType.NUMBER,
+      null,
+      doubleValue,
+      null,
+      null
+    );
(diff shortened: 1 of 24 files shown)


src/main/java/cloud/eppo/api/EppoValue.java
```diff
@@ -136,21 +197,4 @@ public boolean equals(Object otherObject) {
   public int hashCode() {
     return Objects.hash(type, boolValue, doubleValue, stringValue, stringArrayValue);
   }
-
-  /** This method is to allow for Android 21 support; String.join was introduced in API 26 */
-  private static String joinStringArray(List<String> stringArray) {
```
hbordersTwitch, line 141: Moved this into `EppoValueType`

src/main/java/cloud/eppo/ufc/dto/EppoValueType.java
```diff
@@ -1,9 +1,198 @@
...
+        @Nullable Double doubleValue,
+        @Nullable String stringValue,
+        @Nullable List<String> stringArrayValue) {
+      throwIfNull(stringArrayValue, "stringArray must not be null");
+      return EppoValueType.joinStringArray(stringArrayValue);
+    }
+  },
+  ;
+
+  public abstract boolean isNull();
```
hbordersTwitch, line 168: Making these abstract lets us remove the `default` case from `EppoValue.toString()`.

src/main/java/cloud/eppo/ufc/dto/VariationType.java
```diff
@@ -1,20 +1,28 @@
...
-  VariationType(String value) {
+  VariationType(@NotNull String value) {
+    throwIfNull(value, "value must not be null");
     this.value = value;
   }
 
-  public static VariationType fromString(String value) {
-    for (VariationType type : VariationType.values()) {
+  @Nullable
+  public static VariationType fromString(@NotNull String value) {
```
hbordersTwitch, line 24: This was a source of a bug. Code assumed this was method returned not-null, but it can return null

src/main/java/cloud/eppo/ufc/dto/adapters/FlagConfigResponseDeserializer.java
```diff
@@ -84,64 +88,70 @@ public FlagConfigResponse deserialize(JsonParser jp, DeserializationContext ctxt
     return new FlagConfigResponse(flags, banditReferences, dataFormat);
   }
 
-  private FlagConfig deserializeFlag(JsonNode jsonNode) {
-    String key = jsonNode.get("key").asText();
+  @NotNull
+  private FlagConfig deserializeFlag(@NotNull JsonNode jsonNode) {
+    @NotNull final String key = jsonNode.get("key").asText();
```
hbordersTwitch, line 92: Since `jsonNode.asText()` returns an empty string:
> Method that will return a valid String representation of the container value, if the node is a value node (method isValueNode returns true), otherwise empty String.

we can assume that every call with `.asText()` is `@NotNull`

PR #237, opened 2026-05-28
fix: feature/v4 bug surfaced by review

typotter:
## Summary

Fixes all 8 critical findings from the code review of feature/v4

- **CR-01**: `loadConfigurationAsync` could leave `CompletableFuture` permanently incomplete if `fetchAndSaveFromRemoteAsync` threw synchronously, or if graceful mode swallowed an exception without completing the future
- **CR-02**: `remoteFetchFuture`, `configurationFuture`, `initialConfigSet` in `ConfigurationRequestor` now `volatile` for cross-thread visibility
- **CR-03**: `setInitialConfiguration` no longer calls `.join()` inside `synchronized(configurationStore)` — replaced with `thenCompose` to avoid deadlock risk with custom store implementations
- **CR-04**: `EppoConfigurationResponse.getBody()` now returns a defensive copy of the internal `byte[]`
- **CR-05**: `isGracefulMode` in `BaseEppoClient` now `volatile` since it's mutated via public setter but read by async threads
- **SDK-01**: `BanditParametersResponseDeserializer` guards against null/missing JSON fields — skips malformed bandit entries with a warning instead of NPE
- **SDK-02**: `OkHttpEppoClient.buildRequest()` validates `HttpUrl.parse()` result (null check) and wraps build exceptions into the `CompletableFuture` for consistent error handling
- **DTO-01**: All 13 DTO `Default` constructors now enforce `@NotNull` params via `Objects.requireNonNull`

## Test plan

- [x] `./gradlew assemble` passes
- [x] `./gradlew test` passes (all existing tests green)

### DIFF
--- a/eppo-sdk-common/src/main/java/cloud/eppo/ufc/dto/adapters/BanditParametersResponseDeserializer.java
+++ b/eppo-sdk-common/src/main/java/cloud/eppo/ufc/dto/adapters/BanditParametersResponseDeserializer.java
@@ -49,41 +49,60 @@ public class BanditParametersResponseDeserializer
     }
 
     Map<String, BanditParameters> bandits = new HashMap<>();
-    banditsNode
-        .iterator()
-        .forEachRemaining(
-            banditNode -> {
-              String banditKey = banditNode.get("banditKey").asText();
-              String updatedAtStr = banditNode.get("updatedAt").asText();
-              Instant instant = Instant.parse(updatedAtStr);
-              Date updatedAt = Date.from(instant);
-              String modelName = banditNode.get("modelName").asText();
-              String modelVersion = banditNode.get("modelVersion").asText();
-              JsonNode modelDataNode = banditNode.get("modelData");
-              double gamma = modelDataNode.get("gamma").asDouble();
-              double defaultActionScore = modelDataNode.get("defaultActionScore").asDouble();
-              double actionProbabilityFloor =
-                  modelDataNode.get("actionProbabilityFloor").asDouble();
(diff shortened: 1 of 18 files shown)


aarsilv (approved):
👍

eppo-sdk-common/src/main/java/cloud/eppo/ufc/dto/adapters/BanditParametersResponseDeserializer.java
```diff
@@ -49,43 +49,61 @@ public BanditParametersResponse deserialize(
...
+            String modelName = getRequiredText(banditNode, "modelName");
+            String modelVersion = getRequiredText(banditNode, "modelVersion");
+            JsonNode modelDataNode = banditNode.get("modelData");
+            if (modelDataNode == null || !modelDataNode.isObject()) {
+              log.warn("Skipping bandit {}: missing modelData", banditEntry.getKey());
+              return;
+            }
+            double gamma = modelDataNode.get("gamma").asDouble();
+            double defaultActionScore = modelDataNode.get("defaultActionScore").asDouble();
+            double actionProbabilityFloor = modelDataNode.get("actionProbabilityFloor").asDouble();
```
sameerank, line 70: nit: Why not a `getRequiredDouble` (like the earlier `getRequiredText`)? I see that the block that catches the exception already logs the relevant info if the numeric fields are missing (`"gamma"`, `"defaultActionScore"`, `"actionProbabilityFloor"`), but the different validation paths for text vs double feels odd, or could use an explanation in a comment if this was intentional

PR #238, opened 2026-05-29
chore: release versions for common 4.0.0 and framework 0.1.0

typotter:
## Summary

- Bumps `eppo-sdk-common` from `4.0.0-SNAPSHOT` → `4.0.0`
- Bumps `eppo-sdk-framework` from `0.1.0-SNAPSHOT` → `0.1.0`
- Updates README usage example from `3.13.1` → `4.0.0`
- Removes snapshot Maven repository references from release dependency examples in `MIGRATION_GUIDE_v4.md` and `FRAMEWORK_SDK_GUIDE.md`
- Adds `MIGRATION_GUIDE_v4.md` and `FRAMEWORK_SDK_GUIDE.md` to the repo
- Adds `snapshot-release` Makefile target for triggering snapshot publish workflows

## Test plan

- [ ] Verify `./gradlew build` passes
- [ ] Verify `ext.isReleaseVersion` evaluates to `true` for both modules
- [ ] Verify no SNAPSHOT version strings remain in consumer-facing documentation

PR #239, opened 2026-05-29
chore: per-artifact release workflows + v4.0.0 / framework v0.1.0 version bumps

typotter:
## Summary

- Replaces the single `publish-sdk.yml` (triggered by GitHub release events) with two separate `workflow_dispatch` workflows — one per Maven artifact
- Bumps versions for the v4 release: `sdk-common-jvm` → `4.0.0`, `eppo-sdk-framework` → `0.1.0`
- Fixes guide docs and adds `snapshot-release` Makefile target (from prior work on this branch)
- Fixes pre-existing Maven badge URL typo in README

## Release workflows

| Artifact | Workflow | Tag format |
|---|---|---|
| `eppo-sdk-framework` | `publish-framework.yml` | `eppo-sdk-framework-vN.N.N` |
| `sdk-common-jvm` | `publish-common.yml` | `sdk-common-jvm-vN.N.N` |

Each workflow:
1. Verifies the `version` input matches the version declared in `build.gradle`
2. Runs the full test suite
3. Stages and deploys the artifact to Maven Central via JReleaser
4. Creates and pushes an annotated tag **after** successful deploy (avoids orphaned tags on failed publishes)

Triggered via CLI:
```bash
gh workflow run publish-framework.yml --ref main --field version=0.1.0
gh workflow run publish-common.yml --ref main --field version=4.0.0
```

## Test plan

- [ ] Confirm workflows appear in GitHub Actions UI after merge
- [ ] Dry-run by triggering with a version that does NOT match build.gradle — expect "Version mismatch" failure
- [ ] Trigger real release after merge

### DIFF
--- a/.github/workflows/publish-common.yml
+++ b/.github/workflows/publish-common.yml
@@ -63,6 +63,10 @@ jobs:
           ./gradlew :eppo-sdk-common:clean :eppo-sdk-common:publish -Prelease
 
       - name: Deploy to Maven Central
+        # jreleaserDeploy runs from the root project (version = framework version, e.g. 0.1.0).
+        # The root version controls RELEASE vs SNAPSHOT mode only (active = 'RELEASE' when non-SNAPSHOT).
+        # Artifact coordinates come from the staged files in build/staging-deploy, which carry the
+        # correct sdk-common-jvm version (e.g. 4.0.0) — JReleaser does not re-key them off project.version.
         run: ./gradlew jreleaserDeploy
 
       - name: Tag and release
--- a/.github/workflows/publish-common.yml
+++ b/.github/workflows/publish-common.yml
@@ -1,0 +1,74 @@
+name: Publish Common SDK to Maven Central
+
+on:
+  workflow_dispatch:
+    inputs:
+      version:
+        description: 'Version to release (e.g. 4.0.0)'
+        required: true
+
+env:
+  CI: true
+
+jobs:
+  publish:
+    runs-on: ubuntu-latest
+    permissions:
+      contents: write
+    env:
+      JRELEASER_MAVENCENTRAL_USERNAME: ${{ secrets.MAVEN_CENTRAL_TOKEN_USERNAME }}
+      JRELEASER_MAVENCENTRAL_PASSWORD: ${{ secrets.MAVEN_CENTRAL_TOKEN_PASSWORD }}
+      JRELEASER_NEXUS2_SNAPSHOT_DEPLOY_USERNAME: ${{ secrets.MAVEN_CENTRAL_TOKEN_USERNAME }}
+      JRELEASER_NEXUS2_SNAPSHOT_DEPLOY_TOKEN: ${{ secrets.MAVEN_CENTRAL_TOKEN_PASSWORD }}
+      JRELEASER_GPG_PASSPHRASE: ${{ secrets.GPG_PASSPHRASE }}
+      JRELEASER_GPG_PUBLIC_KEY: ${{ vars.GPG_PUBLIC_KEY }}
+      JRELEASER_GPG_SECRET_KEY: ${{ secrets.GPG_PRIVATE_KEY }}
+      JRELEASER_GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+
+      - name: Set up JDK 8
+        uses: actions/setup-java@v4
+        with:
+          java-version: '8'
+          distribution: 'temurin'
+
+      - name: Verify version
+        run: |
+          ACTUAL=$(./gradlew :eppo-sdk-common:properties -q | grep "^version:" | awk '{print $2}')
+          if [ "$ACTUAL" != "${{ inputs.version }}" ]; then
+            echo "Version mismatch: eppo-sdk-common/build.gradle has $ACTUAL, expected ${{ inputs.version }}"
+            exit 1
+          fi
+
+      - name: Download test data
+        run: make test-data
+
+      - name: Test
+        run: ./gradlew check
+
+      - name: Verify JReleaser
+        run: ./gradlew jreleaserConfig --stacktrace
+
+      - name: Stage artifacts
+        # :eppo-sdk-common:clean only cleans the subproject; explicitly remove root staging-deploy
+        # to prevent stale artifacts from prior runs being re-deployed by jreleaserDeploy
+        # Note: checkVersion in root build.gradle gates :publish but not :eppo-sdk-common:publish;
+        # the version-input verification step above covers the same guard for this workflow
+        run: |
+          rm -rf build/staging-deploy
+          ./gradlew :eppo-sdk-common:clean :eppo-sdk-common:publish -Prelease
+
+      - name: Deploy to Maven Central
+        run: ./gradlew jreleaserDeploy
+
+      - name: Tag release
+        # Tag after successful deploy so a failed publish does not leave an orphaned tag
+        run: |
+          git config user.name "github-actions[bot]"
+          git config user.email "github-actions[bot]@users.noreply.github.com"
+          git tag -a "sdk-common-jvm-v${{ inputs.version }}" -m "Release sdk-common-jvm v${{ inputs.version }}"
+          git push origin "sdk-common-jvm-v${{ inputs.version }}"
--- a/.github/workflows/publish-common.yml
+++ b/.github/workflows/publish-common.yml
@@ -65,10 +65,27 @@ jobs:
       - name: Deploy to Maven Central
         run: ./gradlew jreleaserDeploy
 
-      - name: Tag release
-        # Tag after successful deploy so a failed publish does not leave an orphaned tag
+      - name: Tag and release
+        # Tag and create GitHub release after successful deploy — avoids orphaned tags on failed publishes
+        # Changelog covers commits touching eppo-sdk-common/ since the last sdk-common-jvm-v* tag
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         run: |
           git config user.name "github-actions[bot]"
           git config user.email "github-actions[bot]@users.noreply.github.com"
-          git tag -a "sdk-common-jvm-v${{ inputs.version }}" -m "Release sdk-common-jvm v${{ inputs.version }}"
-          git push origin "sdk-common-jvm-v${{ inputs.version }}"
+
+          TAG="sdk-common-jvm-v${{ inputs.version }}"
+          git tag -a "$TAG" -m "Release sdk-common-jvm v${{ inputs.version }}"
+          git push origin "$TAG"
+
+          LAST_TAG=$(git tag -l "sdk-common-jvm-v*" --sort=-version:refname | sed -n '2p')
+          if [ -n "$LAST_TAG" ]; then
+            RANGE="${LAST_TAG}..HEAD"
+          else
+            RANGE="HEAD"
+          fi
+          NOTES=$(git log "$RANGE" --oneline -- eppo-sdk-common/ | sed 's/^/- /')
+
+          gh release create "$TAG" \
+            --title "sdk-common-jvm v${{ inputs.version }}" \
+            --notes "${NOTES:-No changes recorded}"
--- a/.github/workflows/publish-framework.yml
+++ b/.github/workflows/publish-framework.yml
@@ -60,10 +60,28 @@ jobs:
       - name: Deploy to Maven Central
         run: ./gradlew jreleaserDeploy
 
-      - name: Tag release
-        # Tag after successful deploy so a failed publish does not leave an orphaned tag
+      - name: Tag and release
+        # Tag and create GitHub release after successful deploy — avoids orphaned tags on failed publishes
+        # Changelog covers commits touching the root project since the last eppo-sdk-framework-v* tag,
+        # excluding the eppo-sdk-common subproject directory
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         run: |
           git config user.name "github-actions[bot]"
           git config user.email "github-actions[bot]@users.noreply.github.com"
-          git tag -a "eppo-sdk-framework-v${{ inputs.version }}" -m "Release eppo-sdk-framework v${{ inputs.version }}"
-          git push origin "eppo-sdk-framework-v${{ inputs.version }}"
+
+          TAG="eppo-sdk-framework-v${{ inputs.version }}"
+          git tag -a "$TAG" -m "Release eppo-sdk-framework v${{ inputs.version }}"
+          git push origin "$TAG"
+
+          LAST_TAG=$(git tag -l "eppo-sdk-framework-v*" --sort=-version:refname | sed -n '2p')
+          if [ -n "$LAST_TAG" ]; then
+            RANGE="${LAST_TAG}..HEAD"
+          else
+            RANGE="HEAD"
+          fi
+          NOTES=$(git log "$RANGE" --oneline -- . ':(exclude)eppo-sdk-common/' | sed 's/^/- /')
+
+          gh release create "$TAG" \
+            --title "eppo-sdk-framework v${{ inputs.version }}" \
+            --notes "${NOTES:-No changes recorded}"
--- a/.github/workflows/publish-framework.yml
+++ b/.github/workflows/publish-framework.yml
@@ -1,8 +1,11 @@
-name: Publish release SDK to the Maven Central Repository
+name: Publish Framework SDK to Maven Central
 
 on:
-  release:
-    types: [published]
+  workflow_dispatch:
+    inputs:
+      version:
+        description: 'Version to release (e.g. 0.1.0)'
+        required: true
 
 env:
   CI: true
@@ -10,6 +13,8 @@ env:
 jobs:
   publish:
     runs-on: ubuntu-latest
+    permissions:
+      contents: write
     env:
       JRELEASER_MAVENCENTRAL_USERNAME: ${{ secrets.MAVEN_CENTRAL_TOKEN_USERNAME }}
       JRELEASER_MAVENCENTRAL_PASSWORD: ${{ secrets.MAVEN_CENTRAL_TOKEN_PASSWORD }}
@@ -31,6 +36,14 @@ jobs:
           java-version: '8'
           distribution: 'temurin'
 
+      - name: Verify version
+        run: |
+          ACTUAL=$(./gradlew :properties -q | grep "^version:" | awk '{print $2}')
+          if [ "$ACTUAL" != "${{ inputs.version }}" ]; then
+            echo "Version mismatch: build.gradle has $ACTUAL, expected ${{ inputs.version }}"
+            exit 1
+          fi
+
       - name: Download test data
         run: make test-data
 
@@ -41,7 +54,16 @@ jobs:
         run: ./gradlew jreleaserConfig --stacktrace
 
       - name: Stage artifacts
-        run: ./gradlew clean publish -Prelease
+        # :clean clears root build/ including build/staging-deploy, ensuring no stale artifacts are deployed
+        run: ./gradlew :clean :publish -Prelease
 
       - name: Deploy to Maven Central
         run: ./gradlew jreleaserDeploy
+
+      - name: Tag release
+        # Tag after successful deploy so a failed publish does not leave an orphaned tag
+        run: |
+          git config user.name "github-actions[bot]"
+          git config user.email "github-actions[bot]@users.noreply.github.com"
+          git tag -a "eppo-sdk-framework-v${{ inputs.version }}" -m "Release eppo-sdk-framework v${{ inputs.version }}"
+          git push origin "eppo-sdk-framework-v${{ inputs.version }}"
--- a/FRAMEWORK_SDK_GUIDE.md
+++ b/FRAMEWORK_SDK_GUIDE.md
@@ -1,0 +1,870 @@
+# Eppo SDK Framework Guide
+
+This guide explains how to use the `eppo-sdk-framework` artifact to implement custom HTTP clients and configuration parsers.
+
+**Use this guide if:**
+- You need a custom JSON library (e.g., Gson, Moshi instead of Jackson)
+- You have custom networking requirements (custom SSL, proxies, authentication)
+- You're on a platform where OkHttp or Jackson aren't suitable (e.g., GraalVM native image)
+
+**Don't use this guide if:**
+- You can use the standard `sdk-common-jvm` artifact (it includes ready-to-use implementations)
+
+---
+
+## Table of Contents
+
+1. [Overview](#overview)
+2. [Artifact Configuration](#artifact-configuration)
+3. [Implementing ConfigurationParser](#implementing-configurationparser)
+4. [Implementing EppoConfigurationClient](#implementing-eppoconfigurationclient)
+5. [Complete Example: Gson Implementation](#complete-example-gson-implementation)
+6. [Testing Your Implementation](#testing-your-implementation)
+7. [Reference: sdk-common-jvm Source](#reference-sdk-common-jvm-source)
+
+---
+
+## Overview
+
+The framework provides two interfaces you must implement:
+
+| Interface | Purpose | Input | Output |
+|-----------|---------|-------|--------|
+| `ConfigurationParser<JSONFlagType>` | Parse JSON bytes to DTOs | Raw JSON bytes | Parsed DTO objects |
+| `EppoConfigurationClient` | Fetch configuration from API | Request object | Response with bytes |
+
+### Architecture
+
+```
+┌──────────────────────┐
+│  Your Application    │
+└──────────┬───────────┘
+           │
+    ┌──────▼────────┐
+    │ BaseEppoClient│
+    └──────┬────────┘
+           │
+           ├─────────────────┐
+           │                 │
+    ┌──────▼──────────┐ ┌───▼─────────────────┐
+    │  Your Parser    │ │  Your HTTP Client    │
+    │  Implementation │ │  Implementation      │
+    └─────────────────┘ └──────────────────────┘
+```
+
+---
+
+## Artifact Configuration
+
+### Gradle
+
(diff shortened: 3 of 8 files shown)


build.gradle
```diff
@@ -23,7 +23,7 @@ plugins {
 apply plugin: "com.diffplug.spotless"
 
 group = 'cloud.eppo'
-version = '0.1.0-SNAPSHOT'
+version = '0.1.0'
```
aarsilv, line 26: 🤖 Comment from Claude: **Snapshot publishing is broken once this lands.** `checkVersion` (line ~190) throws `"You cannot specify -Psnapshot with a non-SNAPSHOT version"`, and it's a `dependsOn` of `publish`. With the root version now committed as `0.1.0` (and `eppo-sdk-common` at `4.0.0`), `publish-snapshot.yml` — which runs `./gradlew clean publish -Psnapshot` on any `snapshot/*` push — will fail at the staging step. The new `make snapshot-release` target added in this same PR is the trigger for exactly that flow, so it can never succeed against `main` as committed.\n\nUsual fix: after cutting a release, bump `main` back to the next `-SNAPSHOT` (e.g. `0.1.1-SNAPSHOT` / `4.0.1-SNAPSHOT`). Your own updated README even says step 1 is to "drop the `-SNAPSHOT` suffix," which implies `main` is expected to normally carry a SNAPSHOT version — but this PR leaves it non-SNAPSHOT.

Makefile
```diff
@@ -46,3 +46,11 @@ test-data:
 .PHONY: test
 test: test-data build
 	./gradlew check --no-daemon
+
+## snapshot-release - Push current branch to snapshot/<branch> to trigger snapshot publish workflow.
+.PHONY: snapshot-release
+snapshot-release:
```
aarsilv, line 52: 🤖 Comment from Claude: This target pushes to `snapshot/<branch>`, which triggers `publish-snapshot.yml` → `./gradlew clean publish -Psnapshot`. Because this PR also pins the versions to non-SNAPSHOT (`0.1.0` / `4.0.0`), that workflow will fail `checkVersion` (`-Psnapshot` with a non-SNAPSHOT version). See my note on `build.gradle` line 26 — this target won't produce a successful snapshot until `main` carries `-SNAPSHOT` versions again.

.github/workflows/publish-common.yml
```diff
@@ -0,0 +1,91 @@
...
+        # :eppo-sdk-common:clean only cleans the subproject; explicitly remove root staging-deploy
+        # to prevent stale artifacts from prior runs being re-deployed by jreleaserDeploy
+        # Note: checkVersion in root build.gradle gates :publish but not :eppo-sdk-common:publish;
+        # the version-input verification step above covers the same guard for this workflow
+        run: |
+          rm -rf build/staging-deploy
+          ./gradlew :eppo-sdk-common:clean :eppo-sdk-common:publish -Prelease
+
+      - name: Deploy to Maven Central
+        run: ./gradlew jreleaserDeploy
```
aarsilv, line 70: 🤖 Comment from Claude: There's an implicit ordering dependency between the two release workflows that nothing here enforces. `sdk-common-jvm` `api`-depends on `eppo-sdk-framework`, and its generated POM (see `eppo-sdk-common/build.gradle` `withXml`) declares `eppo-sdk-framework:<root version>` as a compile dependency. If `publish-common.yml` is run before `publish-framework.yml` has released that exact framework version to Maven Central, consumers of `sdk-common-jvm` get an unresolvable transitive dependency.\n\nSince these are independent `workflow_dispatch` triggers, consider either documenting "release framework first, then common" prominently, or adding a guard that checks the framework version is already available on Central before deploying common.

.github/workflows/publish-common.yml
```diff
@@ -0,0 +1,91 @@
...
+            exit 1
+          fi
+
+      - name: Download test data
+        run: make test-data
+
+      - name: Test
+        run: ./gradlew check
+
+      - name: Verify JReleaser
```
aarsilv, line 53: 🤖 Comment from Claude: `jreleaserConfig` / `jreleaserDeploy` run from the root project, whose `version` is `0.1.0` (the framework), while this workflow is staging and deploying `sdk-common-jvm` `4.0.0`. The JReleaser config block lives in the root `build.gradle`, so `project.version` it sees is `0.1.0`. The deploy reads staged files from `build/staging-deploy` (which have the correct `4.0.0` POMs), so the upload itself is likely fine — but please confirm `jreleaserDeploy` doesn't key any validation, bundle naming, or metadata off the `0.1.0` project version when the only staged artifact is `4.0.0`. A mismatch here can fail or mislabel the Central deployment.

README.md
```diff
@@ -13,30 +13,40 @@ or [JVM](https://github.com/Eppo-exp/java-server-sdk) SDKs.
...
+gh workflow run publish-common.yml --ref main --field version=4.0.0
+```
+
+The workflow will:
+- Verify the version in `build.gradle` matches the input
+- Run all tests
+- Sign and publish the artifact to Maven Central
+- Create and push a tag (`eppo-sdk-framework-vN.N.N` or `sdk-common-jvm-vN.N.N`) on successful deploy
+
+Monitor progress at [Actions](../../actions).
```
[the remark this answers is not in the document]
typotter: Fixed — both relative `../../actions` links in README.md replaced with absolute URLs (commit 38bf074).

Makefile
```diff
@@ -46,3 +46,11 @@ test-data:
...
+	$(eval LOCAL_BRANCH := $(shell git rev-parse --abbrev-ref HEAD))
+	@echo "$(INFO)Pushing $(LOCAL_BRANCH) to snapshot/$(LOCAL_BRANCH)$(END)"
+	git push origin HEAD:refs/heads/snapshot/$(LOCAL_BRANCH)
```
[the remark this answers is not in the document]
typotter: Fixed — added a guard that checks for `HEAD` and fails fast with a clear error message before the push (commit 38bf074).

eppo-sdk-common/build.gradle
```diff
@@ -6,7 +6,7 @@ plugins {
 apply plugin: "com.diffplug.spotless"
 
 group = 'cloud.eppo'
-version = '4.0.0-SNAPSHOT'
+version = '4.0.0'
```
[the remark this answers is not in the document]
typotter: Addressed in README.md — step 4 of the release instructions now explicitly requires bumping `main` back to the next `-SNAPSHOT` version after both artifacts are confirmed on Maven Central, with a note that this must happen before snapshot publishing will work again (commits 38bf074 and 67e5694).

build.gradle
typotter: Addressed — README step 4 now documents the post-release SNAPSHOT bump as a required step, with explanation that snapshot publishing will fail until it's done. The PR intentionally leaves versions at `4.0.0`/`0.1.0` since the intent is to cut the release immediately after merge; the SNAPSHOT bump is the follow-up commit.

Makefile
typotter: Acknowledged in README — the instructions now clarify that `make snapshot-release` won't work until the post-release SNAPSHOT bump is made. Also added the detached HEAD guard to the target itself (commit 38bf074).

.github/workflows/publish-common.yml
typotter: Documented — README now has a prominent blockquote in the Steps section requiring framework to be released and confirmed on Maven Central before triggering `publish-common.yml`, with the reason explained (POM embeds framework version at publish time). Also added a note that the SNAPSHOT bump must wait until both artifacts are confirmed to avoid the same problem in reverse (commits 38bf074, 67e5694).

.github/workflows/publish-common.yml
typotter: Confirmed safe and documented. `jreleaserDeploy` uses `active = 'RELEASE'` mode (triggered by root version being non-SNAPSHOT) and deploys whatever is in `build/staging-deploy` by path — it does not re-key artifact coordinates off `project.version`. The staged files from `eppo-sdk-common:publish` carry `artifactId=sdk-common-jvm` and `version=4.0.0` in their POMs; JReleaser uploads them as-is. Added a comment in the workflow explaining this (commit 38bf074).

MIGRATION_GUIDE_v4.md
```diff
@@ -0,0 +1,637 @@
...
+10. [Utils.Base64Codec](#utilsbase64codec)
+11. [Quick Migration Checklist](#quick-migration-checklist)
+12. [Complete Examples](#complete-examples)
+
+---
+
+## Artifacts Overview
+
+### v3.x Structure
+- Single artifact: `cloud.eppo:eppo-sdk-framework` (used for all implementations)
```
aarsilv: 🤖 **Comment from Claude working with Aaron:**

**Suggest dropping this file from the PR.**

Two reasons.

It was removed deliberately. `a37985f` — "chore: remove stale MIGRATION_GUIDE_v4.md and .planning/PROJECT.md", 2026-08-20 — is a standalone commit that does nothing else. It was squash-folded into #248, which is why it looks like collateral. `git merge-base --is-ancestor 4935893 b7e71c6` confirms the guide was in your tree when you removed it. Re-adding it here reverts that call without saying so.

It is also stale against current `main`. `ConfigurationParser` now takes two type parameters and `buildConfig`, and no longer declares `parseBanditParams` (`src/main/java/cloud/eppo/parser/ConfigurationParser.java:16,44`; the method survives on the concrete `eppo-sdk-common/src/main/java/cloud/eppo/JacksonConfigurationParser.java:113`). `BaseEppoClient` takes two type parameters (`src/main/java/cloud/eppo/BaseEppoClient.java:29`). A null `configurationStore` now throws (`src/main/java/cloud/eppo/BaseEppoClient.java:82-84`). All three v4 Java snippets under Complete Examples use stale API shapes: `:529` and `:574` declare a one-parameter `ConfigurationParser`, `:555` a one-parameter `BaseEppoClient`, and `:559` passes null for the store.

This line is also inverted:

> - Single artifact: `cloud.eppo:eppo-sdk-framework` (used for all implementations)

v3 published `sdk-common-jvm`, not `eppo-sdk-framework`. `v3.13.2:build.gradle:76` is `artifactId = 'sdk-common-jvm'`. `eppo-sdk-framework` is new in v4. Line 489 has the same inversion.

If you want the guide back, a separate PR off current main is cleaner. Note that dropping it leaves an existing dangling link on main — `FRAMEWORK_SDK_GUIDE.md:880` points at the absent file. Worth fixing separately either way.

.github/workflows/publish-framework.yml
```diff
@@ -0,0 +1,87 @@
...
+      - name: Set up JDK 8
+        uses: actions/setup-java@v4
+        with:
+          java-version: '8'
+          distribution: 'temurin'
+
+      - name: Verify version
+        run: |
+          ACTUAL=$(./gradlew :properties -q | grep "^version:" | awk '{print $2}')
+          if [ "$ACTUAL" != "${{ inputs.version }}" ]; then
```
aarsilv: 🤖 **Comment from Claude working with Aaron:**

**The version gate can be bypassed by its own input.**

`${{ inputs.version }}` is substituted into the script text before bash parses it. Dispatch with the literal 7 characters `$ACTUAL` and the check renders as:

```
if [ "$ACTUAL" != "$ACTUAL" ]; then
```

That always passes. This is the only guard that compares the dispatch input against the project version, and the job holds `JRELEASER_GPG_SECRET_KEY` and the Maven Central token.

Pass the input through `env:` and validate the shape:

```yaml
      - name: Verify version
        env:
          INPUT_VERSION: ${{ inputs.version }}
        run: |
          set -euo pipefail
          if [[ ! "$INPUT_VERSION" =~ ^[0-9]+\.[0-9]+\.[0-9]+$ ]]; then
            echo "::error::version must be MAJOR.MINOR.PATCH, got '$INPUT_VERSION'"; exit 1
          fi
          ACTUAL=$(./gradlew :properties -q | awk -F': ' '/^version: /{print $2; exit}')
          if [ -z "$ACTUAL" ]; then echo "::error::could not read project version"; exit 1; fi
          if [ "$ACTUAL" != "$INPUT_VERSION" ]; then
            echo "::error::Version mismatch: build.gradle has '$ACTUAL', input was '$INPUT_VERSION'"; exit 1
          fi
```

I ran this. Accepts `4.0.0`; rejects `$ACTUAL`, `4.0.0"; echo PWNED; echo "`, `v4.0.0`, `4.0`, `4.0.0-SNAPSHOT`, empty, and a genuine mismatch.

There are ten such substitutions, five per workflow. Apply the same `env:` indirection at `:43`, `:73`, `:74` and `:86` here, and at `:42`, `:43`, `:81`, `:82`, `:94` in `publish-common.yml`. The regex also removes the `-SNAPSHOT` hole in `publish-common.yml`, where `checkVersion` does not gate `:eppo-sdk-common:publish`.

.github/workflows/publish-framework.yml
```diff
@@ -0,0 +1,87 @@
...
+      version:
+        description: 'Version to release (e.g. 0.1.0)'
+        required: true
+
+env:
+  CI: true
+
+jobs:
+  publish:
+    runs-on: ubuntu-latest
```
aarsilv: 🤖 **Comment from Claude working with Aaron:**

**A release can be cut from any ref that carries this workflow.**

`workflow_dispatch` accepts any `--ref` that contains the workflow, and neither workflow declares an `environment:` or checks `github.ref`. The workflow definition itself comes from the dispatched ref, so an unreviewed branch can publish a signed artifact to Maven Central. The old `release: published` trigger required a separate, auditable act.

```yaml
jobs:
  publish:
    runs-on: ubuntu-latest
    if: github.ref == 'refs/heads/main'
    environment: maven-central-release
    permissions:
      contents: write
```

The `if:` is a guard against mistakes only — someone editing the workflow on their own branch removes it. The enforcement is the environment: in repo settings, restrict its deployment branches to `main`, add a required reviewer, and move the Maven and GPG secrets into it so no other ref can read them.

Credit where due: `permissions: contents: write` is already correct and minimal, and better than the workflow it replaces, which declared none.

build.gradle
aarsilv: 🤖 **Comment from Claude working with Aaron:**

**Two things worth settling before this line ships a real release.** GitHub will not let me anchor outside the diff, so both refer to lines elsewhere in this file.

**The published POM points at a repository that does not exist.** `:128`, `:142`, `:143` and `:144` all say `Eppo-exp/sdk-common-jvm`. That 404s. The repo is `Eppo-exp/sdk-common-jdk`. `eppo-sdk-common/build.gradle:100,114-116` has the same four. Maven Central is immutable, so 4.0.0 would carry a dead SCM URL permanently. `:138` and `eppo-sdk-common/build.gradle:110` also set `email = 'https://geteppo.com'`, which is a URL.

While changing these, the protocols are also stale: `git://` was disabled by GitHub, and the `ssh://` form is missing `git@`.

```groovy
url = 'https://github.com/Eppo-exp/sdk-common-jdk'
connection = 'scm:git:https://github.com/Eppo-exp/sdk-common-jdk.git'
developerConnection = 'scm:git:ssh://git@github.com/Eppo-exp/sdk-common-jdk.git'
```

**There is no Actions release run on the JReleaser version main now pins.** `origin/main:build.gradle:20` reads:

```groovy
id 'org.jreleaser' version '1.25.0' // Stay at 1.18. Version 1.20.0 fails to publish.
```

Dependabot #245 bumped the version and left the warning. Snapshot deploy is proven on 1.25.0. The most recent successful release run was `v3.13.2`, on 1.18.0. I cannot rule out a local run, only an Actions one. Either re-pin to 1.18.0 and add a dependabot `ignore` for `org.jreleaser*`, or accept 1.25.0 and replace the comment with what was verified. A `// Stay at 1.18` comment on a `1.25.0` line is misleading during an incident.

.github/workflows/publish-common.yml
```diff
@@ -0,0 +1,95 @@
...
+        # :eppo-sdk-common:clean only cleans the subproject; explicitly remove root staging-deploy
+        # to prevent stale artifacts from prior runs being re-deployed by jreleaserDeploy
+        # Note: checkVersion in root build.gradle gates :publish but not :eppo-sdk-common:publish;
+        # the version-input verification step above covers the same guard for this workflow
+        run: |
+          rm -rf build/staging-deploy
+          ./gradlew :eppo-sdk-common:clean :eppo-sdk-common:publish -Prelease
+
+      - name: Deploy to Maven Central
+        # jreleaserDeploy runs from the root project (version = framework version, e.g. 0.1.0).
```
aarsilv: 🤖 **Comment from Claude working with Aaron:**

**The root version decides where this deploy goes, and nothing checks it.**

The comment here is accurate: `jreleaserDeploy` runs from the root project. But mode selection is not benign. `build.gradle:167` sets `active = 'RELEASE'` on `mavenCentral`, and `:174` sets `active = 'SNAPSHOT'` on `nexus2`. Both key off the root version.

Follow the README's own steady state: root sits at `0.1.1-SNAPSHOT` between releases (`README.md:51`), and a common-only patch bumps only `eppo-sdk-common/build.gradle` (`README.md:33`). Dispatch `publish-common.yml` with `version=4.0.1`. The gate passes, because it reads only the subproject. Then `mavenCentral` is inactive and `nexus2` is active, so a release-versioned artifact is pushed through the snapshot deployer. Whether that endpoint accepts it or rejects it needs a live run. If it accepts, the workflow goes on to tag and publish a GitHub release for something the Maven Central release repository never received.

`eppo-sdk-common/build.gradle:129` compounds it: the common POM embeds `project(':').version`, so the same state writes a SNAPSHOT framework dependency into a release POM.

Add a root-version assertion to the Verify step:

```yaml
          ROOT=$(./gradlew :properties -q | awk -F': ' '/^version: /{print $2; exit}')
          case "$ROOT" in
            *-SNAPSHOT)
              echo "::error::root version is $ROOT. Releasing sdk-common-jvm now would embed a SNAPSHOT framework dependency in the release POM and route the deploy to the snapshot repository. Release the framework first, or set the root version to an already-published framework release."
              exit 1 ;;
          esac
```

This only rejects a SNAPSHOT root; it does not confirm the framework release is live on Central. For that, add a `curl -fsI https://repo1.maven.org/maven2/cloud/eppo/eppo-sdk-framework/$ROOT/eppo-sdk-framework-$ROOT.pom` check.

.github/workflows/publish-framework.yml
```diff
@@ -0,0 +1,87 @@
...
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          git config user.name "github-actions[bot]"
+          git config user.email "github-actions[bot]@users.noreply.github.com"
+
+          TAG="eppo-sdk-framework-v${{ inputs.version }}"
+          git tag -a "$TAG" -m "Release eppo-sdk-framework v${{ inputs.version }}"
+          git push origin "$TAG"
+
+          LAST_TAG=$(git tag -l "eppo-sdk-framework-v*" --sort=-version:refname | sed -n '2p')
```
aarsilv: 🤖 **Comment from Claude working with Aaron:**

**The previous-tag lookup picks the wrong tag when you release a patch behind a newer minor.**

`sed -n '2p'` assumes the tag just created sorts first. It does not when you release a patch behind a newer minor.

I reproduced it. Tags `sdk-common-jvm-v4.0.0` and `sdk-common-jvm-v4.1.0` exist; release `4.0.1` from a branch off `sdk-common-jvm-v4.0.0`:

```
current logic  -> LAST_TAG=sdk-common-jvm-v4.0.1   notes: (empty)
```

The tag just created is picked as its own predecessor, so the release notes read "No changes recorded" and the fix is invisible.

Resolve from ancestry, before creating the tag:

```bash
LAST_TAG=$(git describe --tags --abbrev=0 --match 'eppo-sdk-framework-v*' HEAD 2>/dev/null || true)
if [ -n "$LAST_TAG" ]; then RANGE="${LAST_TAG}..HEAD"; else RANGE="HEAD"; fi
NOTES=$(git log "$RANGE" --oneline -- . ':(exclude)eppo-sdk-common/' | sed 's/^/- /')
TAG="eppo-sdk-framework-v${INPUT_VERSION}"
git tag -a "$TAG" -m "Release eppo-sdk-framework v${INPUT_VERSION}"
git push origin "$TAG"
```

This uses `$INPUT_VERSION`, so the Tag step needs `INPUT_VERSION: ${{ inputs.version }}` added to its existing `env:` block alongside `GH_TOKEN` — the same indirection as the Verify step. Same shape for `publish-common.yml:85` with the `sdk-common-jvm-v*` match and the `eppo-sdk-common/` pathspec. On the same backport this returns `sdk-common-jvm-v4.0.0` and lists the fix.

.github/workflows/publish-common.yml
```diff
@@ -0,0 +1,95 @@
...
+  workflow_dispatch:
+    inputs:
+      version:
+        description: 'Version to release (e.g. 4.0.0)'
+        required: true
+
+env:
+  CI: true
+
+jobs:
```
aarsilv: 🤖 **Comment from Claude working with Aaron:**

**Consider a concurrency group.**

Two dispatches of the same workflow can run at once. They do not share a filesystem — each job gets its own runner — so staged artifacts are safe. The shared state is remote. Duplicate dispatches for the same ref and version can submit the same coordinates concurrently. One may fail during deployment; if both deploy steps succeed, they can then collide while pushing the tag or creating the release, after deployment has already been submitted.

```yaml
concurrency:
  group: publish-common
  cancel-in-progress: false
```

`cancel-in-progress: false` matters — never kill a run that may already be mid-deploy. Same for `publish-framework.yml` with `group: publish-framework`.

Makefile
```diff
@@ -46,3 +46,15 @@ test-data:
 .PHONY: test
 test: test-data build
 	./gradlew check --no-daemon
+
+## snapshot-release - Push current branch to snapshot/<branch> to trigger snapshot publish workflow.
+.PHONY: snapshot-release
+snapshot-release:
+	$(eval LOCAL_BRANCH := $(shell git rev-parse --abbrev-ref HEAD))
```
aarsilv: 🤖 **Comment from Claude working with Aaron:**

**Take this side of the conflict wholesale.**

`main` gained `snapshot-release` through #238, which was split out of this branch. Your version is main's target plus the detached-HEAD guard, so it is a strict superset and the conflict is only that both sides added the same target.

The guard is a real fix. Without it, in detached HEAD `git rev-parse --abbrev-ref HEAD` returns `HEAD`, so the target pushes a branch literally named `snapshot/HEAD` — and `publish-snapshot.yml` matches `snapshot/*`, so that publishes a signed snapshot from a detached commit.

During `git merge origin/main` on this branch, that is `git checkout --ours -- Makefile`. I tested the direction: on the feature branch, `--ours` keeps the guard and `--theirs` discards it.

FRAMEWORK_SDK_GUIDE.md
```diff
@@ -0,0 +1,880 @@
+# Eppo SDK Framework Guide
```
aarsilv: 🤖 **Comment from Claude working with Aaron:**

**This file is already identical on `main`; it only appears here because the branch is behind.**

`git diff origin/main HEAD -- FRAMEWORK_SDK_GUIDE.md` is empty. It shows as +880 because the three-dot diff runs from a merge base that predates #238, which landed this file on main.

Worth knowing before you rebase: #238 squashed `121c00c` and `ea63361` together, so main holds the corrected text while a rebase replays the uncorrected original. That produces add/add conflicts in this file even though the end state is a no-op. Resolve with `git checkout --ours -- FRAMEWORK_SDK_GUIDE.md`, which keeps main's corrected text.

MIGRATION_GUIDE_v4.md
typotter: Fixed in f3d7453 —  removed from the PR.

.github/workflows/publish-framework.yml
typotter: Fixed in f3d7453 — all `${{ inputs.version }}` interpolations in both workflows now route through an `env: INPUT_VERSION:` binding. Added MAJOR.MINOR.PATCH shape validation that rejects `$ACTUAL`, injection strings, `v`-prefixed, SNAPSHOT, and empty inputs.

.github/workflows/publish-framework.yml
typotter: Fixed in f3d7453 — both jobs now declare `if: github.ref == 'refs/heads/main'` and `environment: maven-central-release`. To complete the enforcement, restrict the `maven-central-release` environment's deployment branches to `main` and add a required reviewer in repo settings; move the Maven Central and GPG secrets there.

build.gradle
typotter: Fixed in f3d7453 — both `build.gradle` files updated: repo URL corrected to `Eppo-exp/sdk-common-jdk`, `git://` protocol replaced with HTTPS, `ssh://` form now includes `git@`, email field changed from a URL to `sdk@geteppo.com`. JReleaser comment updated to reflect that 1.25.0 is in use and snapshot deploys are verified on it (no Actions release run on 1.25.0 yet).

.github/workflows/publish-framework.yml
typotter: Fixed in f3d7453 — both workflows now resolve `LAST_TAG` via `git describe --tags --abbrev=0 --match '...' HEAD` before creating the new tag. The changelog range is computed from ancestry, not sorted tag list, so backport patches get the correct predecessor.

.github/workflows/publish-common.yml
typotter: Fixed in f3d7453 — both workflows now declare `concurrency: group: publish-{framework,common}, cancel-in-progress: false`.

Makefile
typotter: The detached-HEAD guard was already present in this branch from commit 38bf074. The `git merge origin/main` that brought in #238's `snapshot-release` target was clean — our version (superset with the guard) was kept.

.github/workflows/publish-common.yml
typotter: Fixed in 81daabd — two guards added to the Verify step: (1) rejects dispatch if root version is SNAPSHOT, preventing the artifact routing to the snapshot deployer and a SNAPSHOT framework dependency in the release POM; (2) `curl -fsI` against Maven Central confirms `eppo-sdk-framework:${ROOT}` is indexed before proceeding — ensures common is not published against a framework version consumers cannot resolve.

